PR #2, opened 2020-02-11
Emit object

aegger13:
Add the emittance scan python object.
Add test emittance scan
Add tests


lcls_tools/emit_scan/mat_emit_scan_test.py
```diff
@@ -0,0 +1,84 @@
+import sys
+import unittest
+import numpy as np
+from mat_emit_scan import MatEmitScan as MES
+
+FILE = 'test_scan.mat'
```
nneveu: Where did these numbers come from? The log output?

PR #6, opened 2020-10-14
Adding archiver functionality

lisazacarias:


nneveu (commented):
Hi Lisa! A couple of comments:
- on the Python3 branch, I had named this folder data_analysis, I'm wondering if it should be something different all together? Maybe just archiver for all archiver related tools? i.e. I would move or add functions into this class?
- we're trying to do unit test files for all classes. Would you like me to make your test a unit test, or would you have time to do that? 
- I think all the other classes only use camel case for the class name, and not functions in the class. What's your opinion on this?

Thanks for the commit!!! :)

lisazacarias:
I love this so much. Is there a reason you don't have a master unit test file that gets run every build? I can write some unit tests that can get ported over whenever.
Personally I prefer camel case because I find it faster to type haha But it's also how I learned, so I'm clearly biased. I can accommodate whatever standard you want to enforce!

nneveu:
Thanks! :) 
- There is no good reason we don't have a master test file. It would be awesome if you could write/add one! 
- I think you're probably right that it's easier to type, but I'm thinking about all the functions I'd have to rename...sad face..
- And what do you think about a general data_analysis, or 'archiver' folder?

nneveu:
@lisazacarias is this ready for one more review and merge? I forgot where this left off, sorry!

lisazacarias:
@nneveu I definitely also forgot, but I think it's ready for another review!

nneveu (approved):
Merging this instead of version in python3 branch bc supports 2/3. Notes for others, camel case used vs. '_' used in other functions.

nneveu:
@lisazacarias can you [sign your commits](https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/about-commit-signature-verification)? then you this should get merged / I approved.

nneveu (approved):
Thanks!

nneveu (approved):

PR #7, opened 2021-08-04
Created superconducting utility classes

Derikka:


nneveu (approved):
@Derikka I'm approving this commit, but can you push again with a few more comments for each function? Especially in the Linac area, i.e. 01 representing a whole cryomodule, and not the first cavity in a cryomodule? Thanks!

PR #8, opened 2021-08-17
Added comments, rack class, and changed class hierarchy for scLinac.py

Derikka:


nneveu (approved):
Thanks for the comments!

PR #11, opened 2022-03-17
fix imports

nneveu:
First pass at trying to fix the import issues.
Modules should be available after a pip install, for example:
`from lcls_tools.image_processing import image`
`from lcls_tools.devices import scLinac`


lisazacarias (approved):
This looks good to me! As someone who doesn't know much about it. Have you considered using versioneer? I think that's supposed to help with tagging releases

PR #13, opened 2022-03-23
Renaming/adding constants 

lisazacarias:


lcls_tools/devices/scLinac.py
```diff
@@ -62,7 +63,7 @@ def __init__(self, linacName, cryomoduleStringList, cavityClass=Cavity, cryomodu
         """
 
         self.name = linacName
-        self.cryomodules: Dict[str, cryomoduleClass] = {}
+        self.cryomodules: Dict[str, Cryomodule] = {}
```
Derikka, line 66: Line 49 says `self.cavities: Dict[int, cavityClass] = {}` which doesn't match what's written in line 66. Do we want to change Cryomodule to cryomoduleClass to be consistent with line 49?

Derikka (approved):

PR #17, opened 2022-05-06
Massive reorg into common, NC, and SC modules

lisazacarias:


jjanice (approved):
So much work!

nneveu:
I'm starting to review this, so far so good. Goal is to finish review today.

lcls_tools/common/pydm_tools/displayUtils.py
```diff
@@ -0,0 +1,35 @@
+from functools import partial
+
+from PyQt5.QtWidgets import QMessageBox, QWidget
+from qtpy.QtCore import Slot
+
+
+@Slot()
```
nneveu, line 7: What is the context for Slot here? Please add a comment, thx.

lcls_tools/common/pydm_tools/magnet.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class MagnetScreen(Display):
+    def __init__(self, parent=None, args=None):
+        super().__init__(parent, args)
+
+    def ui_filename(self):
+        return 'magnet_template.ui'
+
+    def connectSignals(self, magnet: scLinac.Magnet):
```
nneveu, line 14: Is this sc specific? 'scLinac.Magnet'?

lcls_tools/common/pydm_tools/magnet.py
```diff
@@ -0,0 +1,41 @@
...
+        self.ui.expertButton.macros = ["DEV={dev}".format(dev=magnet.pvprefix[:-1])]
+        self.ui.magnetGroupBox.setTitle('CM{cm} {magnettype}'.format(cm=magnet.cryomodule.name, magnettype=magnet.name))
```
nneveu, line 16: more SC specific things here?

nneveu: cryomodule.name

lcls_tools/common/pyepics_tools/pyepicsUtils.py
```diff
@@ -0,0 +1,4 @@
+EPICS_NO_ALARM_VAL = 0
```
nneveu, line 1: What are these for? I'm not familiar.

lcls_tools/superconducting/scLinacUtils.py
```diff
@@ -0,0 +1,123 @@
...
+
+# these values are based on the list of enum states found by probing {Magnettype}:L{x}B:{cm}85:CTRL
+MAGNET_RESET_VALUE = 10
+MAGNET_ON_VALUE = 11
+MAGNET_OFF_VALUE = 12
+MAGNET_DEGAUSS_VALUE = 13
+MAGNET_TRIM_VALUE = 1
+
+
+class PulseError(Exception):
```
nneveu, line 41: These three functions are all the same inside? I think they may be called in different scenarios?  Maybe there's a way to have a general error func that returns the type instead of many of the same errors w/ different names? We can discuss if that doesn't make sense.

setup.py
```diff
@@ -1,13 +1,12 @@
-from setuptools import setup, find_packages
-import sys
+from setuptools import find_packages, setup
 
 setup(
-    name='lcls-tools',
-    version='0.1.dev0',
```
nneveu, line 6: Version was removed?

nneveu (changes requested):
Why were qe/test_scan.mat files and scLinac deleted? 
I think the latter was just moved w/o git mv. I guess for the mat files it's ok to remove for now, but maybe we should start a tests/regression tests folder for those things. See other comments for small suggestions/questions I had. Thanks, this is great!

nneveu:
Also, fyi for work flow, I think this might have nicely fit into three PRs?
- moving all the common files
- moving making the nc folder
- moving making the sc folder

lcls_tools/common/pydm_tools/magnet.py
lisazacarias: Ah yeah I wasn't sure how to write this annotation considering that a file for NC magnets doesn't exist yet, but both types will be of type "Magnet" and have the same parameters

lcls_tools/common/pyepics_tools/pyepicsUtils.py
lisazacarias: It's for checking a PV's severity (like, whether the PV is red or yellow or purple)

lcls_tools/superconducting/scLinacUtils.py
lisazacarias: Yeah they're thrown in different scenarios and trigger different actions. I couldn't think of a better way to do it but I agree that this was is painful haha I'm not sure how to make it a function without pre defining the errors...

setup.py
lisazacarias: Ah fuck that was when I was trying to fix the reinstallation. I'll put it back

Derikka (approved):

lcls_tools/common/pydm_tools/magnet.py
nneveu: Should we take a stab at a general magnet file that lives in the pydm dir?

setup.py
nneveu: ty!

nneveu (approved):
Approving this to get bulk changes in main. We should definitely look two things in the near term though: 

- general magnet class in pydm folder 
- unified error signal function 

I'll make issues for these

nneveu:
Please do a smaller PR next time, and TY for all the hard/awesome work!!

PR #23, opened 2022-06-28
Image Analysis Electron

rubaiaambia:
Detected image and found the values to replace the old ones. Fixed the bugs that were on image_processing_test.py


nneveu (commented):
Closing PR bc of ipynb.

PR #25, opened 2022-07-14
Updating SSA calibration and cavity RF ramps, loosening loaded Q tolerance

lisazacarias:


lcls_tools/superconducting/scLinac.py
```diff
@@ -32,23 +32,24 @@ def __init__(self, cavity):
         self.currentSlopePV: PV = PV(self.pvPrefix + "SLOPE")
         self.measuredSlopePV: PV = PV(self.pvPrefix + "SLOPE_NEW")
         
-        self.maxdrive_pv: str = self.pvPrefix + "DRV_MAX_REQ"
-        self.drivemax = 1 if self.cavity.cryomodule.isHarmonicLinearizer else 0.8
+        self.maxdrive_setpoint_pv: str = self.pvPrefix + "DRV_MAX_REQ"
+        self.saved_maxdrive_pv: str = self.pvPrefix + "DRV_MAX_SAVE"
+        self.drivemax = 1 if self.cavity.cryomodule.isHarmonicLinearizer else caget(self.saved_maxdrive_pv)
```
s-aderhold, line 37: Should we have a initial/default value for the drivemax of the 1.3s, in case there is no value in the pv for some reason? Or is that unlikely/impossible?

lisazacarias: Hmm. Good point.

s-aderhold (approved):

PR #41, opened 2022-12-10
PV class wrapper with smart get, doing some refactoring

lisazacarias:


lcls_tools/superconducting/scLinac.py
```diff
@@ -209,60 +206,61 @@ def move(self, numSteps: int, maxSteps: int = utils.DEFAULT_STEPPER_MAX_STEPS,
...
-
-            print("Motor moving", datetime.now())
-            sleep(1)
-            if (abs(self.cavity.detune_best_PV.value) > 150000
-                    and self.cavity.freq_stop_pv.value != 400000):
+                        f"Abort requested for {self.cavity.cryomodule.name} cavity {self.cavity.number} stepper tuner")
+            
+            print(f"{self} motor moving", datetime.now())
+            
+            if abs(self.cavity.detune_best_PV.get()) > 150000:
```
s-aderhold, line 256: Do we just keep the chirp range at 400k? or would it make sense to make it smaller again when the cavity gets closer?

lcls_tools/superconducting/scLinac.py
```diff
@@ -603,143 +607,156 @@ def setup_rf(self, desAmp):
...
-                                           upperLimit=utils.CAVITY_SCALE_UPPER_LIMIT,
-                                           pushPV=self.pushCavityScalePV,
-                                           savePV=self.saveCavityScalePV,
-                                           exception=utils.CavityScaleFactorCalibrationError)
-
-        print(f"{self} calibration successful")
-
+        
+        print(f"running {self} cavity characterization")
+        self.cavityCalibrationStartPV.put(1)
```
s-aderhold, line 711: to be consistent, might be good to rename this one to 'cavityCharacterizationStartPV'

s-aderhold (commented):
left a few comments

lcls_tools/superconducting/scLinac.py
lisazacarias: I might just delete this because it depends on whether we're going to resonance or staying at cold landing 🙄

PR #48, opened 2023-03-01
Python 3 migration and adding check for too many motor steps

lisazacarias:


s-aderhold (approved):

lcls_tools/superconducting/scLinac.py
```diff
@@ -651,15 +661,21 @@ def setup_tuning(self, chirp_range=200000):
         self.turnOn()
         sleep(5)
         
-        if self.detune_best_PV.severity == EPICS_INVALID_VAL:
-            raise utils.DetuneError(f"{self} Detune PV invalid. Either expand the chirp"
-                                    " range or use the rack large frequency scan"
-                                    " to find the detune.")
+        self.find_chirp_range(chirp_range)
+    
+    def find_chirp_range(self, chirp_range=200000):
```
lisazacarias, line 666: @s-aderhold can you double check this? I really wanna make sure it makes sense

s-aderhold: In general I would say it makes sense. Not sure how I feel about the *1.25 multiplication. If it runs a couple of times you end up with some pretty weird numbers. I would either check with Sonya if the LLRF system is okay dealing with a chirp range of 488281.25Hz or maybe change it to adding 100kHz each step instead.

lisazacarias: I know that any chirp range is fine as long as it's symmetrical ints. Would casting as an int alleviate the concern? I wanted to allow for smaller steps if that made the resulting detune reading more accurate

s-aderhold: that sounds fine then. Cast it as an int to get rid of possible problems.

s-aderhold (approved):

PR #50, opened 2023-05-18
changing timeplots to archiver timeplots

lisazacarias:
This will enable plots to open with archiver data


s-aderhold (approved):
looks good to me

PR #53, opened 2023-05-26
Major refactor: will break many existing programs that depend on it

lisazacarias:
No longer creating PV objects before they're actually needed, and then using caget instead of pv.get() once the latter fails. This is intended to address long existing performance issues


s-aderhold (approved):
looked at it many times. I think this is good

nneveu (commented):
I did a very quick/cursory review. I think at this point just merge and see what breaks, then add fixes as smaller commits.

Derikka (commented):
So many changes! Looks good to me

PR #57, opened 2023-05-31
Improvements for HLs

lisazacarias:
- Only running 1 SSA calibration with the saved drive max
- Lowering cavity characterization drive level to 10%
- Only tuning HLs to 200Hz detune


lcls_tools/superconducting/scLinac.py
```diff
@@ -65,29 +65,22 @@ def drivemax(self):
...
-            print(f"{self.cavity} SSA Calibration failed with '{e}', retrying")
-            self.calibrate(drivemax - 0.02)
-        
-        except utils.SSACalibrationToleranceError as e:
-            print(f"{self.cavity} SSA Calibration failed with '{e}'")
+        except (utils.SSACalibrationToleranceError, utils.SSACalibrationError) as e:
             if attempt < 3:
-                print(f"retyring {self.cavity} calibration")
-                self.calibrate(drivemax, attempt=attempt + 1)
+                self.calibrate(drivemax, attempt + 1)
```
s-aderhold, line 81: why just run it three times with the same parameters?

lisazacarias: Because sometimes (often) it fails once and then succeeds on a rerun. We ran into this problem last night and it was ROUGH

s-aderhold (approved):

PR #75, opened 2023-07-19
differentiate between HL and non-HL for scalefactor check 

s-aderhold:
checking HL and non-HL cavities against separate scalefactor limits and adjusting the thresholds accordingly


lisazacarias (approved):

lisazacarias:
We should still talk to Andy about why the 1.3 range is so huge

s-aderhold:
> We should still talk to Andy about why the 1.3 range is so huge

Agreed. I just want this hurdle out of the way for using autosetup on the HLs for this startup (and the weekend)

PR #76, opened 2023-07-21
trying to catch HL piezo centering sign error

lisazacarias:


s-aderhold (approved):
Approving to fix this for now. I think long term we will want to tackle this in a way that mirrors the actual functional difference and not just fix the outcome.

PR #77, opened 2023-07-24
fix import in magnet.py

s-aderhold:
fix import of sc_linac_utils that got broken by infrastructure changes


lisazacarias (approved):
Teamwork makes the dream work

PR #97, opened 2023-10-23
Fix flake8 warnings

### LINKED ISSUE #96: Clean up Flake8 Errors/Warnings
The new workflow runs Flake8 against lcls-tools but there are a lot of errors/warnings that were present prior to the workflow being put in-place.

The clean-up of lcls-tools in regards to Flake8 such that the flake8 stage of the workflow passes.

MattKing06:
## Summary of Changes
Most of the changes include reformatting or removing unused variable/imports. 
I have added a `.flake8` file that allows the errors/warnings to be configured. For example, there is an allowable line-length that is configured in that file to be 120 (PyCharm default), as well as ignoring some other warnings.

## To Check
- Check out this branch (`96_fix_flake8_warnings`) in your local repository
- Install flake8 (`python3 -m pip install flake8`) if you do not already have it
- Run `flake8 .` in the lcls-tools repository
- There should be no output from this command.

## Other notes
The first stage of the pipeline should pass after these changes have been merged and we can move onto fixing the unit-tests.

closes #96


lcls_tools/superconducting/scLinac.py
```diff
@@ -1696,7 +1696,8 @@ def __init__(
         stepper_class=StepperTuner,
         piezo_class=Piezo,
     ):
-        # type: (str, Linac, Type[Cavity], Type[Magnet], Type[Rack], bool, Type[SSA], Type[StepperTuner], Type[Piezo]) -> None
+        # _type: (str, Linac, Type[Cavity], Type[Magnet], Type[Rack],
```
lisazacarias, line 1699: The underscore breaks the IDE typecasting. Was it included to be able to add the line break? I would prefer keeping the cast over making the line shorter

lisazacarias (changes requested):

lcls_tools/superconducting/scLinac.py
MattKing06: I was trying to break the line so it'd fit within the 120 length. I didn't want to remove it entirely even though it's commented. Let me see if I can remove the _ and keep the line break :)

lisazacarias: Ah yeah it's a comment, but it's an IDE type hint. I tried putting them directly in the argument list (like, `piezo_class: Type[Piezo] = Piezo` but I got errors so I kept it this way

MattKing06: Added a noqa tag next to this line so flake8 ignores line-length here

lisazacarias (approved):
Thanks for all this hard work!

lcls_tools/common/devices/magnet/magnet_constants.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/local/lcls/package/python/current/bin/python
```
nneveu: Do we still need this file?

nneveu (approved):
Thanks!

nneveu:
🥳 🍾 🎆

PR #107, opened 2023-11-09
Cleanup

nneveu:
Some files/directories cleaned up:

- Moved matlab image file from examples to tests/datasets/images/matlab
- Old sol_calc file moved to new beam/transverse folder
- init files removed as needed
- Old grab images example file removed


nneveu:
Moved sol_calc test file to try to pass checks

MattKing06 (approved):
Minor code changes that are just adjusting imports to reorganization so that's good. 
Always good to delete some old files too!

PR #120, opened 2023-11-22
workflow to build spinx docs and move to github pages

MattKing06:
Sphinx will build and the html pages will be moved over to GH-Pages


nneveu (commented):
We still need to move over old gh-pages text before merging this?

nneveu (approved):
Thank you @MattKing06 for helping us get our workflows in order! I'm excited about the docs living with the code. :tada:

PR #128, opened 2024-01-24
adding too and refactoring screen class

roussel-ryan:
This PR does the following: **These are major changes so I'm certainly looking for feedback on how you want to do this**:
- splits the functionality of the screen class into two, a device object that represents the screen object along with associated properties, and a ScreenProfileMeasurement object that combines the device info with the fitting tool to create a measurement. 
- moves saving images and metadata to measurement class since these specifications are relevant more to the measurement process not the hardware object (I think we can have more discussion about this)
- adds a method to the screen object that can measure a background and save that info to file, incl. methods to shutter the beam
- removes some of the timeout logic because it doesn't seem necessary since the operations are blocking, we could also move this logic to the PV class -- there might be something that I'm missing that you guys identified earlier
- adds a yaml env file

### DIFF
--- a/lcls_tools/common/measurements/profile.py
+++ b/lcls_tools/common/measurements/profile.py
@@ -1,0 +1,169 @@
+import os
+import time
+from datetime import datetime
+from threading import Thread
+from typing import Optional, List, Dict, Any
+
+import h5py
+import numpy as np
+import pandas as pd
+from matplotlib import pyplot as plt
+from numpy import ndarray
+from pydantic import (
+    BaseModel,
+    DirectoryPath, PositiveFloat,
+)
+
+from lcls_tools.common.controls.pyepics.utils import PV
+from lcls_tools.common.data_analysis.fitting_tool import FittingTool
+from lcls_tools.common.devices.screen import Screen
+
+
+class ScreenProfileMeasurement(BaseModel):
+    device: Screen
+    fitting_tool: Optional[FittingTool] = None
+
+    save_location: Optional[DirectoryPath] = None
+    save_data: bool = True
+    extra_pvs: List[PV] = []
+    wait_time_in_seconds: PositiveFloat = 1.0
+    condition_list: List = []
+
+    def measure_profile(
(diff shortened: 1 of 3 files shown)


nneveu:
Can you run flake8 and black? the formatting checks are failing and I'd like to see if the tests pass.

roussel-ryan:
I'd like to get general feedback on the overall structure of what I'm proposing before fixing all of the tests etc if possible, then I'd be happy to fine tune w.r.t. formatting

roussel-ryan:
Sounds good, I'm happy to discuss on Monday

lcls_tools/common/measurements/profile.py
```diff
@@ -0,0 +1,169 @@
...
+                        continue
+
+            # if all conditions are satisfied then add the image and extra pv data to
+            # the list
+            images += [img]
+            scalar_measurements += [result]
+
+        return images, scalar_measurements
+
+    def _generate_new_filename(self, extension: Optional[str] = ".h5") -> str:
```
MattKing06, line 128: I always wondered whether this type of stuff should be brought into it's own submodule i.e. `lcls_tools.common.io` or something. This would allow every file to have the same iso timestamp functionality. It would probably involve passing in a file-name for the user, and the extension is already taken care of.

eloiseyang:
> Can you run flake8 and black? the formatting checks are failing and I'd like to see if the tests pass.

@nneveu We should set up separate workflows for style testing and unit testing. That way the unit tests won't get blocked by a failing style test, as they are right now.

shamin-slac:
> Lots of great work, thanks Ryan! Let's chat before or at the SWS meeting. I think raw measurements should stay with the device class and any post processing should be elsewhere. It's definitely up for discussion though! Happy to hear what @eloiseyang @shamin-slac @MattKing06 @kabanaty think too.

I agree with this. I think measurements and processing should be separate. It also might be worth considering separating the device class that has properties from the methods that interact with EPICS.

nneveu:
Closing because this is out of date now.

PR #129, opened 2024-01-29
create measurement class and io function

roussel-ryan:
Adds abstract classes for file io, measurements


roussel-ryan:
lcls_tools_objects

roussel-ryan:
An example of how a basic image diagnostic would work:
```
class ScreenBeamProfile(Measurement):
    name = "beam_profile"
    device: Screen
    image_processer: Any
    fitting_tool: FittingTool
    fit_profile: bool = True

    n_shots: int = 5
    charge_window: Optional[ChargeWindow] = None

    def measure(self) -> dict:
        images = []
        while len(images) < self.n_shots:
            measurement = self.single_measure()
            if len(measurement):
                images += [self.single_measure()]

        # fit profile if requested
        if self.fit_profile:
            results = []
            for ele in images:
                results += [self.fitting_tool.fit_image(ele["processed_image"])]

            # combine images with profile info
            final_results = {}
        else:
            # concat list elements into a single dict
            final_results = {}

        if self.save_data:
            self.dump_controller.dump_data_to_file(final_results, self)

        return final_results

    def single_measure(self) -> dict:
        # measure profiles
        # get raw data
        raw_image = self.device.image

        # get ICT measurements and return None if not in window
        if self.charge_window is not None:
            if not self.charge_window.in_window():
                return {}

        processed_image = self.image_processer.process(raw_image)
        return {"raw_image": raw_image, "processed_image": processed_image}
```

lcls_tools/common/devices/ict.py
```diff
@@ -0,0 +1,35 @@
+from pydantic import field_validator, SerializeAsAny
+
+from lcls_tools.common.devices.device import Device, PVSet, ControlInformation, Metadata
+from epics import PV
+
+
+class ICTPVSet(PVSet):
+    """
+    The PV interface for screens is not uniform.
```
nneveu, line 9: do you mean ICT here instead of screen?

lcls_tools/common/devices/ict.py
```diff
@@ -0,0 +1,35 @@
+from pydantic import field_validator, SerializeAsAny
+
+from lcls_tools.common.devices.device import Device, PVSet, ControlInformation, Metadata
+from epics import PV
+
+
+class ICTPVSet(PVSet):
+    """
```
nneveu, line 8: we would need to update the reader script to pull ICT in yaml files too. Do you have a target ICT/name I could add?

lcls_tools/common/measurements/measurement.py
```diff
@@ -0,0 +1,16 @@
...
+from typing import Optional
+
+from pydantic import BaseModel, DirectoryPath
+
+
+class Measurement(BaseModel, ABC):
+    name: str
+    save_data: bool = True
+    save_location: Optional[DirectoryPath] = None
+
```
nneveu, line 11: should we have metadata placeholders here that all measurements should have? Area? device locations? rmats?

lcls_tools/common/measurements/measurement.py
```diff
@@ -0,0 +1,16 @@
...
+    @abstractmethod
+    def measure(self) -> dict:
+        """ Implements a measurement and returns a dictionary with the results"""
+        pass
```
nneveu, line 15: Trying to think of what else besides metadata would be common to all measurements? I'm sure there's something.

lcls_tools/common/io.py
```diff
@@ -0,0 +1,19 @@
+from typing import Optional
+
+from lcls_tools.common.measurement.measurement import Measurement
+
+
+class HDF5Dumper:
+    def __init__(self):
```
nneveu, line 7: small preference, but I think I would go for H5reader and H5writer?

lcls_tools/common/measurements/measurement.py
nneveu: Maybe beam path over area...

lcls_tools/common/measurements/beam_charge.py
```diff
@@ -0,0 +1,11 @@
+from lcls_tools.common.devices.ict import ICT
+from lcls_tools.common.measurements.measurement import Measurement
+
+
+class BeamChargeMeasurement(Measurement):
+    name = "beam_charge"
```
nneveu, line 6: I would go for BeamCharge vs. BeamChargeMeasurement?

lcls_tools/common/measurements/beam_charge.py
```diff
@@ -0,0 +1,11 @@
...
+    ict_monitor: ICT
+
+    def measure(self) -> dict:
+        return {"bunch_charge_nC": self.ict_monitor.get_charge()}
+
```
nneveu, line 11: Assign this to a variable, as we'll prob add more things to the dict that gets returned?

nneveu (commented):
Looks like a great start, thanks Ryan!

lcls_tools/common/devices/ict.py
roussel-ryan: no idea

lcls_tools/common/measurements/measurement.py
roussel-ryan: the measurements will contain device objects, which themselves contain the metadata. We can add a property to the measurement object that collects this info for a user if necessary

lcls_tools/common/measurements/measurement.py
roussel-ryan: I imagine something will come up as we implement more measurements

lcls_tools/common/measurements/beam_charge.py
roussel-ryan: sure thats fine, but should be done in a device specific PR. This is just a simple placeholder

roussel-ryan:
Keep in mind that the scope of this PR is specific to constructing the overall framework (abstract classes) and changes can be easily made to the implementations of specific functions. Any implementations shown here are just examples to understand how the classes interact with one another

lcls_tools/common/measurements/beam_charge.py
nneveu: If it's something we know we will do, I lean towards just adding it in now for style/format guidance to future contributors. At some point it will be hard to keep track of things.

roussel-ryan:
> Let's use `NotImplemented` instead of pass. I think it's fine to have this as a framework, but if there are things we know need to add, I'd prefer to take the extra time to do it now. That way it stands as a better example. I can commit some of the changes.

Ok thats fine if you want to mostly finish the ICT measurement example, I'm holding off on adding device specific code (ie. screen code, @phys-cgarnier's projection fitting code) until this PR is merged.

roussel-ryan:
I decided against specifying a separate multi-shot wrapper class since each type of measurement might have different ways of effectively measuring and calculating statistics

nneveu (approved):

nneveu:
Note, will add an issue to update yaml files to have a real ICT example.

Issue #132, opened 2024-02-08
Allow SC stepper motors to handle a negative maximum step size


lisazacarias:
**Describe the bug**
If the user passes in a negative max step size, the condition `if abs(numSteps) <= maxSteps` will always fail

**To Reproduce**
1. Call `StepperTuner.move()` with a negative `maxSteps`

**Expected behavior**
If the magnitude of the requested step size is > the _magnitude_ of the max, only step max steps in the appropriate direction as needed

**Screenshots**
![image](https://github.com/slaclab/lcls-tools/assets/13070175/eb021e8c-3437-442d-8c3d-f121f37509d3)

**Desktop (please complete the following information):**
 - ACR OPI

**Smartphone (please complete the following information):**
N/A

**Additional context**
This was called from the SRF Parking GUI asking for movement by steps, not frequency

PR #133, opened 2024-02-08
Allowing for bypassing detune check during motor move

lisazacarias:
Also addressing issue #132

### DIFF
--- a/lcls_tools/superconducting/scLinac.py
+++ b/lcls_tools/superconducting/scLinac.py
@@ -5,7 +5,7 @@
 ################################################################################
 from datetime import datetime
 from time import sleep
-from typing import Callable, Dict, List, Type, Optional
+from typing import Callable, Dict, List, Optional, Type
 
 from numpy import sign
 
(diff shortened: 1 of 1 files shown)


lcls_tools/superconducting/scLinac.py
```diff
@@ -480,39 +480,48 @@ def move(
...
         :param changeLimits: whether to change the speed and steps
+        :param check_detune: whether to check for valid detune after each move
         :return:
         """
 
         self.check_abort()
+        maxSteps = abs(maxSteps)
 
         if changeLimits:
             # on the off chance that someone tries to write a negative maximum
```
nneveu, line 498: Is a negative no longer a problem?

lcls_tools/superconducting/scLinac.py
```diff
@@ -480,39 +480,48 @@ def move(
...
+                speed,
+                changeLimits=False,
+                check_detune=check_detune,
+            )
 
-    def issueMoveCommand(self, numSteps):
+    def issueMoveCommand(self, numSteps: int, check_detune: bool = True):
         # this is necessary because the tuners for the HLs move the other direction
         if self.cavity.cryomodule.is_harmonic_linearizer:
             numSteps *= -1
```
nneveu, line 527: Maybe this is the answer to my q above. You check here now?

lcls_tools/superconducting/scLinac.py
lisazacarias: This just immediately turns max steps into a positive number regardless of what the user requested

lcls_tools/superconducting/scLinac.py
lisazacarias: No, this is numSteps, which is slightly different (I should probably figure out a way to rewrite this lol). It's the number of steps requested by the user vs the maximum number of steps allowed by the stepper

lcls_tools/superconducting/scLinac.py
nneveu: So this is the max steps _possible_ for all cavities, which is some default in utils.DEFAULT_STEPPER_MAX_STEPS? Is the default sometimes neg?

nneveu: Lisa clarified, it's the maxSteps for one tuner movement.

nneveu (approved):
Thanks!

PR #137, opened 2024-02-21
Giving cavity characterization more time before throwing an error

lisazacarias:
The last global setup led to characterizations that took ~1.5 minutes and erroneously marked them as failed. Also adding missing abort check

### DIFF
--- a/lcls_tools/superconducting/scLinac.py
+++ b/lcls_tools/superconducting/scLinac.py
@@ -1554,7 +1554,7 @@ class Cavity(utils.SCLinacObject):
         if self.characterization_status == utils.CALIBRATION_COMPLETE_VALUE:
             if (datetime.now() - self.characterization_timestamp).total_seconds() > 300:
                 raise utils.CavityQLoadedCalibrationError(
-                    f"{self} characterization took longer than 5 minutes to run"
+                    f"No valid {self} characterization within the last 5 min"
                 )
             self.finish_characterization()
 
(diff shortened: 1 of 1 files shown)


lcls_tools/superconducting/scLinac.py
```diff
@@ -1544,16 +1544,17 @@ def characterize(self):
...
                 datetime.now(),
             )
             sleep(1)
 
         if self.characterization_status == utils.CALIBRATION_COMPLETE_VALUE:
-            if (datetime.now() - self.characterization_timestamp).total_seconds() > 60:
+            if (datetime.now() - self.characterization_timestamp).total_seconds() > 300:
                 raise utils.CavityQLoadedCalibrationError(
-                    f"{self} characterization did not start"
+                    f"{self} characterization took longer than 5 minutes to run"
```
s-aderhold, line 1557: waiting for 5min without any feedback on progress makes me guess that people will start saying 'the GUI is frozen'. Maybe 2 or 3minutes? Or can we think of any way to convey that things are actually happening?

lisazacarias: Well the thing is that right now it's just waiting indefinitely (not even just 5 minutes but however long it takes for characterization to run) in the `while self.characterization_running` section. This chunk is just to check that we for sure had a successful characterization within the last 5 minutes, in case it actually didn't start

s-aderhold: Oh right. Maybe the better wording then would be 'no valid characterization within the last 5 min'?

s-aderhold (approved):

PR #140, opened 2024-03-27
Handling failure case when no microseconds provided

lisazacarias:


s-aderhold (approved):
Thanks for fixing this!

PR #149, opened 2024-05-07
adding property for offline cavities

s-aderhold:


lcls_tools/superconducting/sc_linac.py
```diff
@@ -1217,6 +1217,10 @@ def hw_mode(self):
     @property
     def is_online(self) -> bool:
         return self.hw_mode == utils.HW_MODE_ONLINE_VALUE
+
```
lisazacarias, line 1220: I think the linter is unhappy that this line has white space

lisazacarias (approved):

PR #150, opened 2024-05-07
Update README.md

s-aderhold:
added note about linting with flake8


nneveu (approved):
Thanks for adding this note!

PR #166, opened 2024-06-06
Update python-package.yml

nneveu:
Indenting python versions in python-package.yml. Current syntax is breaking linter.


nneveu:
Reviewed w/ Matt on zoom.

PR #177, opened 2024-07-25
Rework as a module, kept simple class, added yaml and output directories.

williamColocho:
Reworked structure by making it a module, adding helper functions outside of the class and adding yaml and output directories.


williamColocho:
Will create a new one

PR #180, opened 2024-08-05
Add z location as property

### LINKED ISSUE #179: Add Z-location as property
**Is your feature request related to a problem? Please describe.**
The Z-location is confusingly called sum_l.  It would be nice to alias sum_l to z_location or something similar.

MattKing06:
## Summary of Changes
Added a property in the `Device` class (`./lcls_tools/common/devices/device.py`) for `z_location` that returns the value of the `sum_l_meters` property so all devices that inherit from it get the property.

Added unit test to ensure that `Device.z_location` and `Device.sum_l_meters` are equal

## To Test

Try running the below script. To try with alternative devices you will have to find a machine area that has those kinds of devices.

```python
from lcls_tools.common.devices.reader import create_area
L1B = create_area("L1B")
# access magnet collection
magnet_names = L1B.magnet_collection.device_names
# pick an arbitrary magnet
magnet = L1B.magnets[magnet_names[0]]
assert(magnet.sum_l_meters == magnet.z_location)
```

## Additional Notes
Currently we only return z co-ordinate, it may be needed in the future to return x/y values too in the case that two components have the same z-location (two beamlines separated at the dipole)

closes #179


nneveu (approved):
Thanks for adding this!

nneveu:
closes #179

PR #187, opened 2024-08-23
Ideas for how to add Effective Length to YAML

phys-cgarnier:
I had attached comments to the code describing the changes I think I should make and where I should make them.


phys-cgarnier:
It looked like magnet.py already had a length property so I didn't make any edits to that code.

lcls_tools/common/devices/yaml/UNDH.yaml
nneveu, line 1: Make length values in YAMLs an integer, also maybe check for empty values? Would we want to put a 0 or something else if there is no effective length?

lcls_tools/common/devices/yaml/metadata.py
```diff
@@ -1,18 +1,20 @@
...
     # }
-    if magnet_names:
-        raise NotImplementedError(
-            "No method of getting additional metadata for magnets."
-        )
-    return {}
+    if magnet_names and method:
+        # Add any additional metadata fields here
+        additional_fields = ['Element', 'Effective Length (m)']
+        device_elements = method(magnet_names, additional_fields)
```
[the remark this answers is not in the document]
phys-cgarnier: As discussed via slack, the field was changed to 'l_eff'

lcls_tools/common/devices/yaml/UNDH.yaml
phys-cgarnier: I made them a float is that alright. 
I also set a fill value of zero for entries in the csv with values ''.

nneveu (changes requested):
@phys-cgarnier sorry to nit pick a little here...can we reduce the number of decimals to the millimeter? I don't think microns is necessary to track on these. Also, the PR is still in draft. I think after this change we can merge quickly. Thanks so much!!

lcls_tools/common/devices/yaml/GUNB.yaml
```diff
@@ -22,6 +22,7 @@ magnets:
       - SC_STMO
       - SC_STXI
       - SC_SXR
+      l_eff: 0.0
```
nneveu, line 25: This should be the same length as CQ02B? Not sure why this one is returning 0.0. We should look into this.

lcls_tools/common/devices/yaml/GUNB.yaml
```diff
@@ -97,6 +100,7 @@ magnets:
...
```
nneveu, line 103: Another 0.0?

lcls_tools/common/devices/yaml/GUNB.yaml
```diff
@@ -147,6 +152,7 @@ magnets:
...
```
nneveu, line 155: Same 0.0

lcls_tools/common/devices/yaml/GUNB.yaml
nneveu, line 1: Lots of 0.0 lengths in this file, we need to check what is going on here.

lcls_tools/common/devices/yaml/HTR.yaml
```diff
@@ -447,6 +464,7 @@ magnets:
       - SC_STMO
       - SC_STXI
       - SC_SXR
+      l_eff: 0.0
```
nneveu, line 467: 0.0 again

lcls_tools/common/devices/yaml/HTR.yaml
nneveu, line 1: More zeros in this file as well.

lcls_tools/common/devices/yaml/L0.yaml
```diff
@@ -112,6 +116,7 @@ magnets:
       - CU_SFTH
       - CU_SPEC
       - CU_SXR
+      l_eff: 0.0
```
nneveu, line 119: Zeros for correctors?

lcls_tools/common/devices/yaml/L0B.yaml
nneveu, line 1: zeros for correctors

nneveu (changes requested):
Lots of zeros in many sections. This might be the right answer in some cases, but we need to double check this.

lcls_tools/common/devices/yaml/L0B.yaml
phys-cgarnier: Hi I am back from leave and will look into this now :)

lcls_tools/common/devices/yaml/GUNB.yaml
phys-cgarnier: I spot checked on CQ01B it is  and the entry is "" in the GUNB.yaml
Which in my code is set to return 0.0.
I will check a few more of these, however I am observing a lot of "" fields in he lcls_elements.csv

PR #189, opened 2024-09-04
Updating wire_test and renaming wire variables

kabanaty:
Updating wire_test to include more tests.  Renamed many wire variables for clarity, e.g. instead of "initialized" (status) and "initialize" (action), have "initialize_status" and "initialize".


lcls_tools/common/devices/yaml/DL1.yaml
```diff
@@ -606,10 +606,10 @@ screens:
   OTR3:
     controls_information:
       PVs:
-        image: OTRS:IN20:621:Image:ArrayData
+        image: OTRS:IN20:621:IMAGE
         n_bits: OTRS:IN20:621:N_OF_BITS
```
nneveu, line 610: OTR2 and 3 seemed to switch camera PVs here? Did you update/ double check this?

lcls_tools/common/devices/yaml/generate.py
```diff
@@ -134,14 +134,17 @@ def _construct_pv_list_from_control_system_name(
...
+                search_term, field = search_term.split(".")
             # End of the PV name is implied in search_term
             try:
                 pv_list = meme.names.list_pvs(name + ":" + search_term, sort_by="z")
                 # We expect to have ZERO or ONE result returned from meme
                 if pv_list != list():
                     if len(pv_list) == 1:
                         # get the pv out of the results
-                        pv = pv_list[0]
+                        pv = f"{pv_list[0]}.{field}" if field else pv_list[0]
```
nneveu, line 147: Is field strictly evaluated as `False` in this case (`str()`)? Maybe we can set field to `False` initially?

nneveu: do we need the string initialization here?

tests/unit_tests/lcls_tools/common/devices/test_wire.py
nneveu, line 1: you can run coverage to check how much code your test covers: https://coverage.readthedocs.io/en/7.3.2/

nneveu (approved):
Only a few minor comments from me. Nothing to prevent merging, thanks!!

nneveu:
You'll need to resolve the yaml conflicts, but should be able to merge after that.

lcls_tools/common/devices/yaml/DL1.yaml
nneveu: Can you pleases write a bug issue on this and reference the code lines, etc. Thanks in advance.

PR #193, opened 2024-09-18
Moving kmod_to_bdes and bdes_to_kmod from bmad_modeling to ../model

williamColocho:


examples/modeling_test.py
```diff
@@ -17,10 +15,10 @@ def tc(cmd):
...
 
 # Instantiate a modeling class with mod.BmadModeling(<beam path>, <'ACT','DES', or 'ARCHIVE'>
-bm = mod.BmadModeling("sc_sxr", "DES")
-bm.date_time = "2024-04-02T09:00:00.000000-08:00"
+bm = mod.BmadModeling('sc_sxr', 'DES')
+bm.date_time = '2024-04-02T09:00:00.000000-08:00'
 
-# See ./lcls_tools/common/data/bmad_modeling/yaml/outkeys.yml
+# See ./lcls_tools/common/data_analysis/bmad_modeling/yaml/outkeys.yml
 # for list of outputs provided by get_output()
```
nneveu, line 22: Sorry in advance for this...I updated data_analysis to be data. That might be why you had so many conflicts. Looks like your PR is trying to put it back to data_analysis. You can probably fix most of these quickly w/ a replace all.

nneveu (changes requested):

examples/modeling_test.py
williamColocho: I clean up the pull request.  Looks good to me now.  Please have a look.

nneveu (approved):
Thanks!

Issue #194, opened 2024-09-19
Allow sum_l to be validated as None


kabanaty:
**Is your feature request related to a problem? Please describe.**
If a device does not have a sum_l defined in the Oracle database, it cannot be created.

**Describe the solution you'd like**
Validate the sum_l property as Union[float, None] in device.py.

Issue #185, opened 2024-08-14
New device class for BLMs


kabanaty:
**Describe the solution you'd like**
Create a new device class for BLMs (beam loss monitors).  Users should be able to read live data as well as BSA/EDEF data.

MattKing06:
closed by #195

PR #195, opened 2024-09-20
Created LBLM device

kabanaty:
Created LBLM device to address #185.  Also addressed #194 in the process of making new device.


.github/workflows/python-app.yml
nneveu, line 1: Why the new workflow file? Was this meant to replace the old one, or was there some other reason for this?

lcls_tools/common/devices/wire.py
```diff
@@ -93,21 +95,21 @@ def validate_pv_fields(cls, v: str) -> PV:
 
 class WireControlInformation(ControlInformation):
     PVs: SerializeAsAny[WirePVSet]
-    _ctrl_options: SerializeAsAny[Optional[Dict[str, int]]] = dict()
+    # _ctrl_options: SerializeAsAny[Optional[Dict[str, int]]] = dict()
```
nneveu, line 99: Were all the control options removed bc they are in the PVs that Ziyu gave us?

lcls_tools/common/devices/yaml/GUNB.yaml
```diff
@@ -1,3 +1,28 @@
...
+  LBLM00A:
+    controls_information:
+      PVs:
+        bypass: LBLM:GUNB:212:A:FAST_AMP_BYP
+        gain: LBLM:GUNB:212:A:FAST_AMP_GAIN
+        i0_loss: LBLM:GUNB:212:A:I0_LOSS
+      control_name: LBLM:GUNB:212:A
+    metadata:
+      area: GUNB
+      beam_path: []
```
nneveu, line 11: Why no beam paths here?

nneveu (approved):
Thanks for this! I will approve, but please make a new issue for getting the LBLM beam paths populated. It might be tricky since they overlap areas, we might need to think about it a bit.

nneveu:
Please resolve the check failure, then we can merge.

.github/workflows/python-app.yml
kabanaty: I was just trying something out and didn't realize it was going to get added to the PR!

lcls_tools/common/devices/wire.py
kabanaty: Yes, basically

lcls_tools/common/devices/yaml/GUNB.yaml
kabanaty: lcls_elements doesn't have any beam paths, for any LBLM devices unfortunately.

.github/workflows/python-app.yml
eloiseyang: Looks like this new workflow file is the source of the [failing test](https://github.com/slaclab/lcls-tools/actions/runs/10966493672/job/31262451134?pr=195) as of October 8th 2024. I think this workflow should be removed from the PR.

kabanaty: Workflow has been removed from the PR

PR #197, opened 2024-09-25
Add analysis functions for emittance measurement scan data

dylanmkennedy:
Add emittance/ folder to data_analysis, containing functions for computing emittance and bmag from quadrupole measurement scan data. emittance/analysis.py contains top-level functions for users to analyze quad scan data, while emittance/beam_dynamics.py contains lower-level functions for tasks such as building quadrupole transport matrices, twiss parameter propagation, and beam matrix reconstruction through nonlinear fitting.


nneveu (commented):
Two minor comments: 

- Can you please pull in the changes from main? 
- Can you consider splitting basic matrix calcs from emittance/bmag calcs. 

I'll submit a longer review later ...

lcls_tools/common/data/emittance/matrix_calcs.py
nneveu, line 1: These functions are not specific to emittance calcs. This should be moved to a more general location. Then we don't need a folder for emittance, and we can just have an 'emittance.py' file.

lcls_tools/common/data_analysis/emittance/beam_dynamics.py
```diff
@@ -0,0 +1,264 @@
+import numpy as np
+from scipy.optimize import minimize
+import importlib.util
+
+
+def compute_bmag(beam_matrix, rmat, beta0, alpha0):
```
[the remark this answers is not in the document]
nneveu: @dylanmkennedy did you look at William's functions/consider merging the two?

lcls_tools/common/data/emittance/fit.py
```diff
@@ -0,0 +1,189 @@
+import numpy as np
+from scipy.optimize import minimize
+import importlib.util
+from matrix_calcs import propagate_twiss, twiss_transport_mat_from_rmat, build_quad_rmat
+from lcls_tools.common.data.model_general_calcs import bmag_func
+
+
+def compute_emit_bmag(
+    k: np.ndarray,
```
nneveu, line 9: Can we use Pydantic to do some type checking here?

lcls_tools/common/data/emittance/fit.py
```diff
@@ -0,0 +1,189 @@
...
+
+    # prepare the A matrix
+    r11, r12 = total_rmat[..., 0, 0], total_rmat[..., 0, 1]
+    amat = np.stack((r11**2, 2.0 * r11 * r12, r12**2), axis=-1)
+    # amat result (batchshape x nsteps x 3)
+
+    beamsize_squared = np.expand_dims(beamsize_squared, -1)
+
+    def beam_matrix_tuple(params):
+        '''
```
nneveu, line 68: Is it a requirement that this is inside the bmag function? Can we pull this out?

lcls_tools/common/data/emittance/matrix_calcs.py
```diff
@@ -0,0 +1,100 @@
...
+            np.stack((-c * cp, c * sp + cp * s, -s * sp), axis=-1),
+            np.stack((cp**2, -2 * cp * sp, sp**2), axis=-1),
+        ),
+        axis=-2,
+    )  # result shape (batchshape, 3, 3)
+    return result
+
+
+def build_quad_rmat(k: np.ndarray, q_len: float, thin_lens: bool=False):
+    """
```
nneveu, line 51: Can some of of these rmats be pulled from the control system directly?

lcls_tools/common/data/emittance/matrix_calcs.py
```diff
@@ -0,0 +1,100 @@
...
+        thin_lens: boolean specifying whether or not to use thin-lens approximation
+
+    Outputs:
+        rmat: numpy array of shape (*k.shape, 2, 2) containing rmats corresponding to the
+                various focusing strengths given by k
+    """
+
+    if not thin_lens:
+        sqrt_k = np.sqrt(np.abs(k))
+
```
nneveu, line 67: Add some more comments maybe? Or a reference from where this came from?

nneveu (changes requested):
Thanks for adding this! I will do another review after some of the comments are resolved.

lcls_tools/common/data_analysis/emittance/beam_dynamics.py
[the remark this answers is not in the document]
dylanmkennedy: Yes, I have already changed it to import and use William's bmag function instead.

lcls_tools/common/data/emittance/matrix_calcs.py
dylanmkennedy: Where would you like me to put the matrix calcs? model_general_calcs.py?

lcls_tools/common/data/emittance/fit.py
dylanmkennedy: It is not a requirement, however it is a function whose only use-case is within the function where I have defined it. Would you prefer that I move it outside of that function?

lcls_tools/common/data/emittance/matrix_calcs.py
dylanmkennedy: Maybe, I'm not sure. In cases where computation time is a factor, I think it would be preferable to not need to make adjustments to the model/retrieve the rmats from it. The idea for this code is to be able to handle large batches of scan data very efficiently.

lcls_tools/common/data/emittance/fit.py
nneveu: ok, so there's no future case where we might want to use this somewhere else?

lcls_tools/common/data/emittance/matrix_calcs.py
nneveu: yes, let's move them there for now, then make the emittance folder a file instead.

lcls_tools/common/data/emittance/fit.py
dylanmkennedy: Do you want me to make an emittance class (pydantic object) with this function as a class method instead of a standalone function?

lcls_tools/common/data/emittance/fit.py
dylanmkennedy: Correct, it is just a convenience function for translating between fit parameters and the real beam matrix parameters, which is only useful during the fitting. I cannot conceive of a situation in which we would be interested in the transformed parameters outside of the fitting process.

lcls_tools/common/data/emittance/matrix_calcs.py
nneveu: There may be a way to grab and store this info vs. calling it new every time. Let's open an issue for that linked to this PR, and we can merge this for now. I'm more worried about differences in rmats in the models used for other applications.

lcls_tools/common/data/emittance/fit.py
nneveu: Since we're renaming the file to emittance.py, let's hold off on that so you can use this next week as is. Can you write another issue for us to think about reformatting? I don't want a user to have to `import emittance.Emittance`, so we should think about it a little more.

nneveu (changes requested):
After the file rename, moving the functions, checks pass we can probably merge this. Let's just make 2 issues for those follow up items.

lcls_tools/common/data/emittance/fit.py
nneveu: After discussion, let's write an issue on the subject of making this a class or not. Also remove the 'emittance' folder, and put emittance.py in data folder.

nneveu (approved):
Thanks for the updates!

PR #199, opened 2024-10-08
Add Profile Measurement Class

eloiseyang:
This is the last portion of [PR 143](https://github.com/slaclab/lcls-tools/pull/143) that hasn't been added yet. I'm setting up this draft pull request as I work on pulling the measurement class up to speed with the changes to the fitting and image classes.


eloiseyang:
I think the tests should run now. I'll do my own review of this soon since this isn't my code. It would be nice to coordinate with Shamin too, since he's working on another measurement at the same time.

eloiseyang:
Don't be alarmed by the number of lines. The jupyter notebook file contains an image, and Github encodes the image as text, resulting in a huge number of lines that aren't actually relevant.

examples/example_measurement.ipynb
nneveu, line 1: Does this need to be a notebook? I'd prefer not to have python notebooks on the repo if possible.

lcls_tools/common/measurements/measurement.py
```diff
@@ -1,10 +1,23 @@
 from abc import ABC, abstractmethod
 from typing import Optional
-
 from pydantic import BaseModel, DirectoryPath
 
 
 class Measurement(BaseModel, ABC):
+    """
```
nneveu, line 7: Let's try the google style RadiaSoft recommended? https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

lcls_tools/common/measurements/screen_beam_profile_measurement.py
```diff
@@ -0,0 +1,99 @@
...
+    device: Screen
+    image_processor: ImageProcessor
+    fitting_tool: ProjectionFit
+    fit_profile: bool = True
+    # charge_window: Optional[ChargeWindow] = None
+
+    def measure(self, n_shots: int = 1) -> dict:
+        """
+        Measurement function that takes in n_shots as argumentment, where n_shots is the number of
+        image profiles (is this what I want to call it?)
```
nneveu, line 34: argumentment -> argument. I think `images', is the right term here vs. profiles. We're taking n pictures of the beam.

lcls_tools/common/measurements/screen_beam_profile_measurement.py
```diff
@@ -0,0 +1,99 @@
...
+        we would like to measure. Invokes single_measure per shot
+        Then if fit_profile = True, fits the x and y projections of each image using the provided
+        fitting_tool. Results are stored in list of dictionaries where each element of the list
+        is results for the image fitting process stored as a dictionary.
+        ----
+        Currently under work, what do we want to do with the images?
+        Needs dump_controller
```
nneveu, line 41: Do you mean the save images / projections to file here?

lcls_tools/common/measurements/screen_beam_profile_measurement.py
```diff
@@ -0,0 +1,99 @@
+from lcls_tools.common.devices.screen import Screen
+from lcls_tools.common.image.processing import ImageProcessor
+from lcls_tools.common.data.fit.projection import ProjectionFit
+from lcls_tools.common.measurements.measurement import Measurement
+import numpy as np
+
+
+class ScreenBeamProfileMeasurement(Measurement):
```
nneveu, line 8: Still not sure beam is necessary in this name...or screen either?  A Profile could be measured with a wire as well, but was the thought a wire profile measurement would be separate code?

lcls_tools/common/measurements/screen_beam_profile_measurement.py
```diff
@@ -0,0 +1,99 @@
...
+        is results for the image fitting process stored as a dictionary.
+        ----
+        Currently under work, what do we want to do with the images?
+        Needs dump_controller
+
+        """
+        images = []
+        while len(images) < n_shots:
+            measurement = self.single_measure()
+            if len(measurement):
```
nneveu, line 47: If a measurement returns a bad image while len(image) < n_shots, there's no catch here?  i.e. if the 3rd out of 10 image measurement returns nothing, it keeps going but would only return 9 good images? Or maybe I'm reading that wrong?

lcls_tools/common/measurements/screen_beam_profile_measurement.py
```diff
@@ -0,0 +1,99 @@
...
+
+                for key, param in self.fitting_tool.fit_projection(
+                    projection_y
+                ).items():
+                    key_y = key + "_y"
+                    temp[key_y] = param
+                final_results += [temp]
+
+            # what should I do with results now?
+            results["fits"] = final_results
```
nneveu, line 75: We'd want to be able to return and save to file associated w/ the images as you have in the next few lines.

tests/unit_tests/lcls_tools/common/measurements/test_screen_beam_profile.py
```diff
@@ -0,0 +1,119 @@
...
+        self._image = image
+
+
+class TestScreenBeamProfileMeasurement(unittest.TestCase):
+    def __init__(self, methodName: str = "runTest") -> None:
+        super().__init__(methodName)
+        self.test_instantiate_pydantic_objects()
+
+    def create_test_image(self, size: tuple, center: list, radius: int):
+        # make img that is a circle in the center of the image with known
```
nneveu, line 34: I would prefer if we use a test file. That way functionality changes would not impact us. We would know we are always testing against the same image.

tests/unit_tests/lcls_tools/common/measurements/test_screen_beam_profile.py
```diff
@@ -0,0 +1,119 @@
...
+        perform_single_measure = self.screen_beam_profile_measurement.single_measure()
+        self.assertIsInstance(perform_single_measure, dict)
+        assert len(perform_single_measure) == 2
+
+    def test_measure(self):
+        perform_measure = self.screen_beam_profile_measurement.measure()
+        self.assertIsInstance(perform_measure, list)
+        self.assertIsInstance(perform_measure[0], dict)
+        if isinstance(self.gauss_model, GaussianModel):
+            for key, val in perform_measure[0].items():
```
nneveu, line 103: I'm not a fan of this...

nneveu (changes requested):
Thanks for submitting this @eloiseyang! Let's continue the discussion in comments.

nneveu:
Also, can you please pull in changes from main? Thanks

examples/example_measurement.ipynb
eloiseyang: I honestly didn't look at it. I can remove it.

lcls_tools/common/measurements/screen_beam_profile_measurement.py
eloiseyang: I think this file was written with a screen profile in mind, but it could be generalized.

lcls_tools/common/measurements/screen_beam_profile_measurement.py
eloiseyang: As the code is written right now, it will keep trying to grab an image until it has n images, regardless of the result. I assume this would only happen if something went wrong in CA. How are we handling CA errors in devices right now?

lcls_tools/common/measurements/screen_beam_profile_measurement.py
eloiseyang: We're using HDF5 to save, right? Is there an example of how we're saving within LCLS-Tools right now?

tests/unit_tests/lcls_tools/common/measurements/test_screen_beam_profile.py
eloiseyang: Should we just remove the value checks? Regression checking the fitting algorithm could be a more complicated task, and this code doesn't really accomplish that I think.

lcls_tools/common/measurements/screen_beam_profile_measurement.py
```diff
@@ -0,0 +1,99 @@
...
+        fitting_tool. Results are stored in list of dictionaries where each element of the list
+        is results for the image fitting process stored as a dictionary.
+        ----
+        Currently under work, what do we want to do with the images?
+        Needs dump_controller
+
+        """
+        images = []
+        while len(images) < n_shots:
+            measurement = self.single_measure()
```
eloiseyang, line 46: I would rather let this function take measurements as an input. The user of this function will probably want to store and manipulate the images on their own.

lcls_tools/common/measurements/screen_beam_profile_measurement.py
```diff
@@ -0,0 +1,99 @@
...
+            if len(measurement):
+                images += [measurement]
+
+        results = {"images": images}
+        if self.fit_profile:
```
eloiseyang, line 51: I'd like to remove this if statement. This should just be a wrapper for the fitting function.

lcls_tools/common/measurements/screen_beam_profile_measurement.py
```diff
@@ -0,0 +1,99 @@
...
+            final_results = []
+            for i, ele in enumerate(images):
+                temp = {}
+                temp["raw_image"] = ele["raw_image"]
+                temp["processed_image"] = ele["processed_image"]
+
+                projection_x = np.array(np.sum(ele["processed_image"], axis=0))
+                projection_y = np.array(np.sum(ele["processed_image"], axis=1))
+
+                for key, param in self.fitting_tool.fit_projection(
```
eloiseyang, line 61: I don't love this syntax.

lcls_tools/common/measurements/screen_beam_profile_measurement.py
```diff
@@ -0,0 +1,99 @@
...
+            results["fits"] = final_results
+
+        # no attribute dump controller
+        if self.save_data:
+            pass
+            # self.dump_controller.dump_data_to_file(final_results, self)
+
+        return final_results
+
+    def single_measure(self) -> dict:
```
eloiseyang, line 84: I don't think this function should be in this file.

lcls_tools/common/measurements/screen_beam_profile_measurement.py
nneveu: Ah yes, I see what you mean. My holiday brain...
Were not doing any sophisticated error handling. A few places with PV checks and attribute errors. 
So it would cause problems if there was an epics error returned/we need to look into that in future PRs.

tests/unit_tests/lcls_tools/common/measurements/test_screen_beam_profile.py
nneveu: Sure, let's go with that and write an issue.

eloiseyang (commented):
Looking at this PR again, I don't want to merge it. The Screen Beam Profile Measurement class has a lot of overlap with the "Profile Monitor GUI Model" class I will eventually include in LCLS-Tools. I think we can put more thought into a class that grabs, processes, and fits data all in one place. That kind of class should probably be broken up into multiple data classes as well.

I'd like to preserve this code in an example script. It's a good demonstration of how the projection fit class works.

lcls_tools/common/measurements/screen_beam_profile_measurement.py
nneveu: There's a [`_write_image_to_hdf5'](https://github.com/slaclab/lcls-tools/blob/main/lcls_tools/common/devices/screen.py) function in the screen class. Although this will need some more thought and will get cleaned up in the demo RadiaSoft is planning.

lcls_tools/common/measurements/screen_beam_profile_measurement.py
robnagler: I think there's an impedance mismatch in the way the function is defined and the implementation, and the extra code for stuff that isn't yet implemented or doesn't actually happen (single_measure returning an empty dict).

Below I changed the function definition to yield a dict per image. An iterator in this case has a number of advantages, e.g. memory efficiency:

```py
    def take_and_fit_images(self, n_shots: int = 1) -> dict:
        """Get `n_shots` from the screen and fit the processed images

        Args:
            n_shots (int): number of valid images to take [1]
        Yields:
            dict: `raw_image`, `processed_image`, and `fits`
        """

        def _fit(processed, axis):
            return self.fitting_tool_fit.fit_projection(
                np.array(np.sum(processed, axis=axis))
            )

        for _ in range(n_shots):
            rv = self.single_measure()
            rv["fits"] = dict(
                {k: _fit(rv["processed_image"], v) for k, v in (("x", 0), ("y", 1))}
            )
            yield rv
```

If the shots need to be all taken all at once (without fitting in between each shot), then the higher level code should pass in the shots (per @eloiseyang's suggestion) controlling the speed, and this function could remain an iterator and would be even simpler:

tests/unit_tests/lcls_tools/common/measurements/test_screen_beam_profile.py
```diff
@@ -0,0 +1,119 @@
...
+                if distance < radius:
+                    image[y, x] = 1
+        return image
+
+    def test_instantiate_pydantic_objects(self):
+        # creating image processing class
+        self.radius = 50
+        self.size = (800, 800)
+        self.center = [400, 400]
+        self.extent = [300, 300]
```
robnagler, line 50: Many of these are unused(?). With this many unnecessary intermediate variables (why on self?), this function is hard to read. Put the mocked objects in a separate function so the intent (test) of this function is clear.

tests/unit_tests/lcls_tools/common/measurements/test_screen_beam_profile.py
```diff
@@ -0,0 +1,119 @@
...
+        self.assertIsInstance(perform_single_measure, dict)
+        assert len(perform_single_measure) == 2
+
+    def test_measure(self):
+        perform_measure = self.screen_beam_profile_measurement.measure()
+        self.assertIsInstance(perform_measure, list)
+        self.assertIsInstance(perform_measure[0], dict)
+        if isinstance(self.gauss_model, GaussianModel):
+            for key, val in perform_measure[0].items():
+                if key == "amplitude_x":
```
robnagler, line 104: Ideally, have these be exact values. Read and write the files to a file (e.g. json or yaml) and them diff that file with an expected output. You can use [`ndiff`] if the numbers are fuzzy. However, testing the fitting function itself is outside the scope of this module so inserting a blank image, for example, would yield a precise value after the fit (0). You would only need to test, say, mean_x and mean_y that they were both 0.

robnagler (commented):
Here are my comments. Hopefully they are useful.

roussel-ryan:
@eloiseyang is this PR outdated now?

eloiseyang:
@roussel-ryan yes, I believe your PR accomplished what this PR was trying to. I'll close it now.

PR #207, opened 2024-11-15
Remove unnecessary beamline argument from emittance measurement class

dylanmkennedy:
Removes the (unused) beamline argument.


nneveu (approved):
Minor change, approving this.

PR #213, opened 2024-11-26
bugfixes from run

roussel-ryan:
Fixes the following bugs:
- image reshaping incorrect
- improper validation of rmats
- incorrect attribute reference for design_twiss
- uses dict output from compute_emit_bmag

### DIFF
--- a/lcls_tools/common/measurements/emittance_measurement.py
+++ b/lcls_tools/common/measurements/emittance_measurement.py
@@ -94,7 +94,7 @@ class QuadScanEmittance(Measurement):
             twiss_design=twiss_betas_alphas,
         )
 
-        results = results.update({
+        results.update({
             "x_rms": self.beam_sizes["x_rms"],
             "y_rms": self.beam_sizes["y_rms"]
         })


roussel-ryan:
@dylanmkennedy can you verify the correct shape for rmat validation?

lcls_tools/common/devices/screen.py
```diff
@@ -73,7 +73,7 @@ def image(self) -> np.ndarray:
         the camera associated with this screen
         """
         return self.controls_information.PVs.image.get(as_numpy=True).reshape(
-            self.n_rows, self.n_columns
+            self.n_columns, self.n_rows
```
nneveu, line 76: How did this mistake get caught? I think I'm remembering this might not be the case for all cameras. Will approve for now and check with Courtney.

lcls_tools/common/measurements/emittance_measurement.py
```diff
@@ -85,20 +86,18 @@ def measure(self):
...
             beamsize_squared=beamsize_squared,
             q_len=magnet_length,
             rmat=rmat,
             twiss_design=twiss_betas_alphas,
         )
 
-        results = {
-            "emittance": emittance,
-            "BMAG": bmag,
+        results = results.update({
```
nneveu, line 97: Do we need the `results =` here if we are updating?

nneveu (approved):

lcls_tools/common/measurements/emittance_measurement.py
robnagler: That actually breaks things, because update returns None.

roussel-ryan: thanks for catching that

PR #214, opened 2024-11-27
Image refactoring

roussel-ryan:
refactor image processing and fitting class to be more general / easier to debug
- creates a base ImageFit abstract class that contains image validation and allows for subclasses to modify the image fitting method
- implements the ImageProjectionFit subclass that uses the ProjectionFit class to fit the x/y projections
- moves all image processing into a single call inside the ImageProcessing class
- replaces object properties with explicit calls to functions that do computation
- makes MethodBase a pydantic class that also contains `use_priors` field to specify the use of priors instead of using a field in ProjectionFit to specify this (defaults to false)


nneveu:
Theres a similar PR draft open on this? Do you want to combine?

roussel-ryan:
Are you referring to #199 ?

nneveu:
I was thinking of #128. Maybe that's too old enough we want to just close it?

roussel-ryan:
Yes, that PR is out of date, we should close it

lcls_tools/common/image/fit.py
```diff
@@ -0,0 +1,164 @@
...
+                bounding_box_penalty = self._calculate_bounding_box_penalty(
+                    image, centroid, sizes,
+                )
+                fit_results["bb_penalty"] = bounding_box_penalty
+
+                # set results to none if the beam extends beyond the roi
+                # and the bounding box constraint is active
+                if bounding_box_penalty > 0 and self.apply_bounding_box_constraint:
+                    fit_results["centroid"] = np.array((np.Nan, np.Nan))
+                    fit_results["rms_sizes"] = np.array((np.Nan, np.Nan))
```
eloiseyang, line 86: Our current practice is to perform scans manually. I could see a value like the bounding box penalty being useful for a human user, but a human user would probably want to preserve the beam parameters, regardless of the quality of the fit. Either we should think more about what the penalty value communicates for a human user, or move the code that parses good and bad data elsewhere.

lcls_tools/common/image/fit.py
```diff
@@ -0,0 +1,164 @@
...
+    def _fit_image(self, image: ndarray) -> dict:
+        x_projection = np.array(np.sum(image, axis=0))
+        y_projection = np.array(np.sum(image, axis=1))
+
+        proj_fit = ProjectionFit(model=self.projection_fit_method)
+
+        x_parameters = proj_fit.fit_projection(x_projection)
+        y_parameters = proj_fit.fit_projection(y_projection)
+
+        results = {
```
eloiseyang, line 157: I've mentioned this a couple times before, in this PR and in others, that I think we should move away from dictionaries of parameters. It's hard to validate what they actually contain as a user. It doesn't have to be fixed in this PR, I think we should come to a solution as a group.

eloiseyang (commented):
Thanks for the PR. My main question for everyone is whether or not the bounding box penalty is useful in the general case. I think LCLS-Tools should be a general code base for physics and HLA applications that integrates easily with ML applications, but shouldn't include automation in its foundation. In my view, base class `ImageFit` should just package fitting with the relevant output parameters for a fit, and validation can occur in a class extending `ImageFit`.

If you have other thoughts, let me know. I'm open to discussion. Thanks again!

lcls_tools/common/frontend/plotting/image.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def plot_image_projection_fit(fit_object, results, n_stds=3):
+    """
+    plot image and projection data for validation
+    """
+    fig, ax = plt.subplots(3, 1)
+    fig.set_size_inches(4, 9)
+
+    image = results["processed_image"]
```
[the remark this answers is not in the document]
roussel-ryan: part of the goal of this function is to easily validate that the image analysis proceeded successfully and to debug any issues. Given the stochastic nature of the fitting process it would be useful to examine the results of fitting in a convenient way to debug / verify proper fitting

roussel-ryan:
thanks for the comments @eloiseyang, I'm working on removing the bounding box stuff now and will update the PR tomorrow morning

eloiseyang (approved):
Thanks for the changes, looks good to me.

Looks like a test in test_sc_cryomodule.py is failing, though you didn't change that one in a way that would make the test fail. It looks like the test here picks a random cryomodule to verify. It probably passed once by chance, but is failing again. I'll take a look myself and let you know.

Edit:

I found the issue, there's two unexpected leading whitespaces in dictionary keys in the test [tests/unit_tests/lcls_tools/superconducting/test_sc_cryomodule.py](https://github.com/slaclab/lcls-tools/pull/214/files/3e3cab1214ec990f27a0ee59ab97a5a613348721#diff-e8a0be63782258fd112cfe683a8d507d8b0661b31a0fcddc765296f9daa11538).

tests/unit_tests/lcls_tools/superconducting/test_sc_cryomodule.py
```diff
@@ -6,13 +6,13 @@
...
     def test_is_harmonic_linearizer_true(self):
-        hl = MACHINE.cryomodules[f"H{randint(1,2)}"]
+        hl = MACHINE.cryomodules[f"H{randint(1, 2)}"]
         self.assertTrue(
             hl.is_harmonic_linearizer, msg=f"{hl} is_harmonic_linearizer is not true"
         )
 
     def test_is_harmonic_linearizer_false(self):
-        cm = MACHINE.cryomodules[f"{randint(1, 35):02d}"]
+        cm = MACHINE.cryomodules[f"{randint(1, 35): 02d}"]
```
eloiseyang, line 15: There shouldn't be a space between the colon and the number here. This changes the format string from
`01` to ` 01`. The dictionary `MACHINE.cryomodules` doesn't expect a leading space in its key names.

roussel-ryan: ok I addressed it, lets see if the tests pass

tests/unit_tests/lcls_tools/superconducting/test_sc_cryomodule.py
```diff
@@ -22,5 +22,5 @@ def test_pv_prefix(self):
         self.assertEqual(cm.pv_prefix, "ACCL:L0B:0100:")
 
     def test_num_cavities(self):
-        cm = MACHINE.cryomodules[f"{randint(1, 35):02d}"]
+        cm = MACHINE.cryomodules[f"{randint(1, 35): 02d}"]
```
eloiseyang: Sorry, there's one more down the line I didn't point out. I think the test will pass after this change.

PR #216, opened 2024-12-06
Magnet length branch

phys-cgarnier:
Updated magnet length branch.

### DIFF
--- a/lcls_tools/common/devices/yaml/generate.py
+++ b/lcls_tools/common/devices/yaml/generate.py
@@ -286,7 +286,7 @@ class YAMLGenerator:
         if basic_magnet_data:
 
             magnet_names = [key for key in basic_magnet_data.keys()]
-            additional_metadata_data = get_magnet_metadata(magnet_names,self.extract_metadata_by_device_names)
+            additional_metadata_data = get_magnet_metadata(magnet_names, self.extract_metadata_by_device_names)
             # should be structured {MAD-NAME : {field_name : value, field_name_2 : value}, ... }
             additional_controls_data = get_magnet_controls_information()
 
--- a/lcls_tools/common/devices/yaml/generate.py
+++ b/lcls_tools/common/devices/yaml/generate.py
@@ -15,7 +15,6 @@ from lcls_tools.common.devices.yaml.controls_information import (
     get_wire_controls_information,
     get_lblm_controls_information,
 )
-import pprint
 
 class YAMLGenerator:
     def __init__(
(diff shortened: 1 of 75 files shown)


lcls_tools/common/devices/yaml/metadata.py
```diff
@@ -1,18 +1,29 @@
...
     #  mag-name-2 : {metadata-field-1 : value-1, metadata-field-2 : value-2},
     #  ...
     # }
-    if magnet_names:
-        raise NotImplementedError(
-            "No method of getting additional metadata for magnets."
-        )
-    return {}
+    if magnet_names and method:
+        # Add any additional metadata fields here
```
nneveu, line 12: What would method be here?

tests/unit_tests/lcls_tools/common/devices/test_magnet.py
```diff
@@ -132,14 +132,18 @@ def test_tol(self) -> None:
...
-        self.assertEqual(self.magnet.length, 0.05)
+        """Test effective length float validation"""
+        # SOL1B -> l_eff: 0.086
+        self.assertEqual(self.magnet.l_eff, .086)
+        # might need validate_assignment = True in magnet class
+        # with self.assertRaises(ValidationError):
+        # self.magnet.l_eff = "a"
+        # some trouble with this.... need to change config later. its not forcing
+        # revalidation after reassignment since type is optional I believe.
+        self.magnet.l_eff = 1.0
```
nneveu, line 143: This doesn't have to prevent merging, but I think some of these tests can be restructured. We shouldn't assign then check the value? Maybe we can compare object value vs. what's in the csv table instead.

lcls_tools/common/devices/yaml/generate.py
```diff
@@ -409,3 +415,30 @@ def extract_lblms(self, area: Union[str, List[str]] = ["HTR"]):
...
+        required_fields=Optional[List[str]]
+    ):
+        # TODO: try not to call filter elements so many times as it parses csv
+        if required_fields:
+            elements = self._filter_elements_by_fields(
+                required_fields=required_fields
+            )
+        else:
+            elements = self._filter_elements_by_fields(
+                required_fields=self._required_fields
```
nneveu, line 431: Sorry, I'm not following the logic here? Maybe a Friday thing. So there is no case where something in `self._required_fields` would be in `required_fields`?

nneveu (approved):
I'm approving because the yaml files look ok. We should rethink this test and not sure followed all the metadata logic.

lcls_tools/common/devices/yaml/metadata.py
phys-cgarnier: Method is the method you want to pass for parsing the lcls_elements.csv file.
Since generate.py had already parsed the csv file for other elements I figured we didn't need to duplicate the code here. Idk what your thoughts on this are.

tests/unit_tests/lcls_tools/common/devices/test_magnet.py
nneveu: Can you write this up in an issue? @phys-cgarnier

PR #221, opened 2025-01-24
Adds h5py dump

pluflou:


lcls_tools/common/data/saver.py
```diff
@@ -0,0 +1,86 @@
...
+
+        dt = h5py.string_dtype(encoding=self.string_dtype)
+        def recursive_save(d, f):
+            for key, val in d.items():
+                if key == 'attrs':
+                    f.attrs.update(val)
+                elif isinstance(val, dict):
+                    group = f.create_group(key, track_order=True)
+                    recursive_save(val, group)
+                else:
```
roussel-ryan, line 50: can we keep the elif isinstance statements going for the other types, incl. string and ndarrays? Better to be explicit. Maybe just call str() on types that do not match anything

pluflou: you mean like this? If we call str() on the else statement, then the numpy data structures won't be saved as numpy types
```python
                elif isinstance(val, dict):
                    group = f.create_group(key, track_order=True)
                    recursive_save(val, group)
                elif isinstance(val, str):
                    f.create_dataset(key, data=val, dtype=dt, track_order=True)
                else:
                    f.create_dataset(key, data=val, track_order=True)
```

roussel-ryan: yes, but add an isinstance(val, np.ndarray) block in there

pluflou: How about the updated code?

```python
        ...

        self.supported_types = (bool, int, float, np.ndarray, np.integer, np.floating)

        ...
                elif isinstance(val, self.supported_types):
                    f.create_dataset(key, data=val, track_order=True)
                elif isinstance(val, str):
                    # specify string dtype to avoid issues with encodings
                    f.create_dataset(key, data=val, dtype=dt, track_order=True)
                else:
                    f.create_dataset(key, data=str(val), track_order=True)
```

roussel-ryan:
@pluflou this would be a good opportunity to merge in the version we have been using

pluflou:
@roussel-ryan do we want this in both packages? I can copy the updated class form ml-tto and mark this as ready for review

roussel-ryan:
yes, lets update the h5 dumping based on ml-tto but I would do a PR to the main branch, not the serialization branch

pluflou:
@eloiseyang @roussel-ryan The class is now up to date with the ml-tto version. Previously I had a `save_data` flag in `ScreenBeamProfileMeasurement` that saves the results to an h5 file  if it's set to true. I removed it until we discuss. Is this something that we want to add in the measurement classes here (screen profile and/or emittance)?

pluflou:
@eloiseyang I added pandas to the requirements since I wrote the `H5Saver` to handle pandas dataframes as well. Is this addition ok? Alternatively, I could check for dataframes in the input only if pandas is installed, and this would save dataframes as strings in the H5 files if they are passed and pandas wasn't installed. We typically use and save dataframes during the ML/Xopt runs that we use lcls-tools for. It could also be useful for a wider range of use-cases that people might encounter while handling data.

lcls_tools/common/data/saver.py
```diff
@@ -0,0 +1,182 @@
...
+        Returns
+        -------
+        dict
+            The dictionary loaded from the file.
+        """
+
+        def recursive_load(f):
+            d = {"attrs": dict(f.attrs)} if f.attrs else {}
+            for key, val in f.items():
+                if isinstance(val, h5py.Group):
```
eloiseyang, line 161: Just to elaborate on my previous comment a little more, h5py groups do two things here. Either they create pandas dataframes or dictionaries by default. We could also write a group that gets reconstructed as a list, which would be a nice way to reconstruct "{key}/{i}" type groups.

I would really recommend that we shrink this PR down. It's difficult to recover data from its printed string form programmatically, and the right way to approach those cases is to expand the code to dump and recover them appropriately, which would add a lot of bulk to this already long PR. Therefore, those parts can be cut.

I think having a generalized way to dump and load data is a good idea in the long run, but we can iterate on our current approach to groups to be general.

lcls_tools/common/data/saver.py
```diff
@@ -0,0 +1,182 @@
...
+                        "pandas_type" in val.attrs
+                        and val.attrs["pandas_type"] == "dataframe"
+                    ):
+                        # Load DataFrame from group
+                        columns = val.attrs["columns"]
+                        data = {}
+                        for col in columns:
+                            data[col] = val[col][:]
+                        d[key] = pd.DataFrame(data)
+                    else:
```
eloiseyang, line 172: We already use a `pandas_type` attribute above to tell the loader what type we're dealing with. Maybe we should add an attribute for dictionaries too?

tests/unit_tests/lcls_tools/common/data/test_saver.py
```diff
@@ -0,0 +1,174 @@
...
+            "a": np.nan,
+            "b": np.inf,
+            "c": -np.inf,
+            "d": [np.nan, np.inf, -np.inf],
+            "e": [np.nan, np.inf, -np.inf, 1.0],
+            "f": [np.nan, np.inf, -np.inf, "a"],
+            "g": {"a": np.nan, "b": np.inf, "c": -np.inf},
+            "h": "np.Nan",
+            "i": np.array((1.0, 2.0), dtype="O"),
+        }
```
eloiseyang, line 26: Should we add assert statements here like in `test_special_values` below?

tests/unit_tests/lcls_tools/common/data/test_saver.py
```diff
@@ -0,0 +1,174 @@
...
+            raw_images=images,
+            processed_images=processed_images,
+            rms_sizes=rms_sizes or None,
+            centroids=centroids or None,
+            total_intensities=total_intensities or None,
+            metadata={"info": "test"},
+        )
+
+        # Dump to H5
+        result_dict = result.model_dump()
```
eloiseyang, line 59: We already test to make sure that a saved dictionary has the same keys as the original when reloaded. Presumably screen beam measurement already tests to make sure the result has the right attributes on output. So long as both of those things are tested separately, this test doesn't test anything new.

tests/unit_tests/lcls_tools/common/data/test_saver.py
```diff
@@ -0,0 +1,174 @@
...
+        saver.dump(data, "test_nested.h5")
+        loaded_data = saver.load("test_nested.h5")
+        os.remove("test_nested.h5")
+
+        assert data["nested_dict"] == loaded_data["nested_dict"]
+        for i in range(len(data["nested_list"])):
+            # lists of lists are saved as dicts
+            # here the lists are saved as nd.arrays
+            assert np.array_equal(
+                data["nested_list"][i], loaded_data["nested_list"][f"{i}"]
```
eloiseyang, line 150: Here's an example of the pattern I'd like to avoid. If we can add an attribute that tells the loader that we're dealing with a pandas dataframe, then there's no reason we can't do the same for a list.

eloiseyang (changes requested):
Hi all, thanks for the PR. I read this code as a generalized approach to dumping and recovering Python objects, like pandas dataframes, ndarrays, and dictionaries. I think that's a good idea, and will be useful for everyone when it's implemented.

The trouble I have with this code is that its approach to different types is inconsistent. A HDF5 group is assumed to be a dictionary by default. Sometimes a group is a Pandas dataframe if it has the appropriate attributes associated with it. Dictionaries and Pandas dataframes are easily dumped and recovered.

Sometimes a group is a list, like when there's nested lists, but these can't be recovered because there's no attribute that handles the reconstruction of lists. When the code dumps a heterogeneous list, it converts the list to a string, which the loader can't recover the data from. The list-group approach in nested lists could be used to solve this problem, but it isn't.

Lists of dictionaries are handled, but tuples and NDArrays of dictionaries are not. Pandas dataframes get to know what type of object they are, but tuples don't. Dictionaries only kind of get to know because that's the default case.

Right now, the code can dump and fully recover scalars, homogenous lists of scalar data, dictionaries, nested dictionaries, and pandas dataframes. I would suggest we reduce the code to only handle these cases that be dumped and recovered in their original form. A generalized way to dump and recover Python objects will definitely be useful, but we should make sure that our code can fully deliver on its promises. We don't want to be scrambling to recover a heterogeneous list of enums, booleans, strings, and doubles from a printed string two years down the line, for instance.

The idea is good and the testing suite is robust. This will be an iterative process, and we don't need to create something that's completely general to start. The first step to getting this merged is reducing it down to the parts of the code that deliver on its promise to save and recover python objects. I'm definitely willing to help iterate on this as well once the initial merge is done. Thank again, let me know if you have questions.

pluflou:
Thank you @eloiseyang for your thorough review! I am wrapping up some other things right now. I plan to address your comments sometime later this week/early next week.

nneveu:
Thank you both for the work on this PR! I second most of the comments Eloise made here. It seems like this PR is doing more than it needs to in the first pass. Maybe some of the if/else-ing can be converted to try/except. Main point though, with out clouding up Eloise's comments, is I would like to see this reduced to the most commonly used case or two. I think we should pick a save convention (or 2-3) and stick with it/enforce it for now. I'm open to discussing if others disagree.

pluflou:
Thank you all for your patience. Just checking in--I got wrapped up in other things and I likely won't get to this until mid/end of May.

pluflou:
Hi all, apologies for the delay on this PR. I have reworked the module to simplify things and only handle the data types/structure that we are dumping in `slac-tools` measurements (I mainly looked at the emittance measurement result dump, including the metadata). I still have to test it a bit more and go over these PR comments more closely to make sure things went in the right direction. I will ping you again when it's ready for a final review. Mainly now it supports:
- scalars (float, bool, int)
- strings
- np.ndarray
- lists (of scalars, nested, and heterogeous)
- tuples
- None types
- pandas dataframes
- dictionaries
- pathlib.PosixPath

If does not support and should raise an implementation error for:
- heterogenous np.ndarrays
- tuples of dictionaries
- np.ndarrays of dictionaries
- lists of dictionaries

tests/unit_tests/lcls_tools/common/data/test_saver.py
pluflou: I mainly added it as a final test that all the info we need/care about here are in fact saved and recovered and that nothing was missed

eloise-nebula (approved):
Sorry for the delay. Thanks for responding to my comments. As I mentioned in our meeting earlier this week, the format we save data in could change with further development.

PR #222, opened 2025-01-27
modularize emittance measurement class

roussel-ryan:
- moves the basic quadrupole scan code to a class method such that the class can be subclassed to use different scanning methods in the future


roussel-ryan:
@eloiseyang is it possible to take a look at why the tests are failing? It looks like a problem with meme install from github

eloiseyang:
Hey @roussel-ryan, I just talked to Matt Gibbs about it. He said he recently changed the meme repo visibility from external to internal. I'm looking into how we might clone an internal repo from within our git actions.

lcls_tools/common/measurements/emittance_measurement.py
```diff
@@ -152,6 +150,12 @@ def measure(self):
 
         return results
 
+    def perform_beamsize_measurements(self):
```
nneveu, line 153: Is this general enough to be called perform beam size measurement? Assumes scanning one quad? Or is the plan to expand this later?

moellep:
For the meme import problem - it looks like it is only used in model_general_calcs.get_optics() which doesn't get called from tests, so a short term work-around could be to move the meme import into that method:
```python

#from meme.model import Model

#...

def get_optics(magnet: Magnet, measurement: Measurement) -> Dict:
    #...
    from meme.model import Model
    model = Model(magnet.metadata.area)
```

eloiseyang:
Hi @roussel-ryan, I fixed the meme issue with PR #224. If you merge with main the tests should start running again.

lcls_tools/common/measurements/emittance_measurement.py
eloiseyang: I agree the name is reasonable. Once we start writing multiquad scans we may want to differentiate between single quad scans, multi, or whatever else we want. I think it's reasonable to call the base class a "single quad scan".

eloiseyang (approved):
Looks good, I think it's ready to merge. If you could merge into main to make the tests run before we merge for a last time that would be great. Thank you!

roussel-ryan:
Still failing the superconducting test, but everything else seems to be working so I will merge

PR #223, opened 2025-01-27
Move image processing from fit.py to screen_profile.py

dylanmkennedy:
Moves the image_processor object and the image processing step from fit.py to screen_profile.py. fit.py will receive the already processed image, and the image_processor will be more easily accessed during screen_profile creation.


nneveu:
If you merge with main, Eloise's last PR should fix your test problem.

PR #229, opened 2025-01-31
Update and cleanup package

pluflou:
- modernized package with `pyproject.toml`
    - added optional deps for `meme` to handle tests locally vs in forks vs on main repo GHA
    - added package versioning from git using `setuptools-scm`
- cleaned up github actions workflows
    - renamed jobs for clarity
    - added option to run jobs on forks w/o `meme`. `meme` dependent tests will be skipped if a fork is running them. A PR from a fork will fail still.
    - added summary output for GHA test results
- updated README 
- fixed a few typos

Notes:
1. With this structure, the dependencies go in `pyproject.toml`(as unpinned as possible), so the library can be installed without running into many issues resolving version constraints. If we want to pin versions for deploying on certain machines, the `requirements.txt` should be used with exact version pinning (equivalent to an `environment.yml` file with conda). For building and releasing (e.g. to PyPI), the `pyproject.toml` is used.
2. As Paul mentioned, the current tests don't use `meme`. So unless we anticipate future tests needing it, the github actions workaround here and the changes to `test_emittance_measurement.py` might be an overkill and his solution would work best.


requirements.txt
```diff
@@ -12,4 +11,4 @@ sphinx
 sphinx_rtd_theme
 myst_parser
 scikit-learn
-meme @ git+ssh://git@github.com/slaclab/meme.git
+meme @ git+https://github.com/slaclab/meme.git
```
eloiseyang, line 14: I'm not able to pip install from my laptop using the url `https://github.com/slaclab/meme.git`, but I can install from `ssh://git@github.com/slaclab/meme.git`. The https url prompts for a username and password, then fails even if I enter the correct credentials. I think we should keep the ssh url.

eloiseyang (commented):
Hi Sara, thanks for the PR, everything looks good. It took me a moment to get used to the .toml file, but I think I understand it now. I made one comment on using an https vs ssh url in requirements.txt.

requirements.txt
pluflou: Oh interesting. I can change it back. Can you install locally with `pip install ".[meme]"` or does that fail as well?

eloiseyang: Good catch, I have the same issue there. If you try to git clone from an https url git says:
```
remote: Support for password authentication was removed on August 13, 2021.
```
I believe ssh is more general.

pluflou: I can change it if you prefer ssh, but I believe both are valid. Your issue is because you are using a password to authenticate with git, and not a personal access token (see [here](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens)). If you set that up, both methods should work.

lisazacarias (approved):

PR #232, opened 2025-02-14
Formalize output of emittance and screen measurements

roussel-ryan:
Major Changes
----------------
Modifies the output of ScreenBeamProfileMeasurement and QuadScanEmittance measurement classes to be pydantic data structures, also adds a utility function to plot results from the quadrupole scan emittance measurement.

Minor changes
----------------
Reformats some docstrings and code based on suggestions from @robnagler


lcls_tools/common/measurements/emittance_measurement.py
```diff
@@ -1,51 +1,144 @@
...
+        Returns
+        -------
+        tuple
+            The best BMAG value and corresponding pv value for the quadrupole.
+
+        """
+        if self.bmag is None:
+            raise ValueError("BMAG values are not available for this measurement")
+
+        if not isinstance(mode, BMAGMode):
```
robnagler, line 98: Encapsulate in BMAGMode, e.g.

```py
    @classmethod
    def from_any(cls, value):
        def _members():
            return ", ".join((m.name for m in cls))

        try:
            if isinstance(value, cls):
		return value
            if isinstance(value, str):
                return cls[value.upper()]
            if isinstance(value, int):
                return cls(value)
        except (ValueError, KeyError):
            pass
        raise ValueError(f"invalid {cls.__name__}={value} must be one of: {_members()}")
```

robnagler (approved):
I made a suggestion which you can take or leave. I have given enough feedback.

roussel-ryan:
@robnagler thank you for the suggestion. I've added it in the code

PR #236, opened 2025-02-20
Pydantic lcls base model

roussel-ryan:
-  globally modifies pydantic acceptance of arbitrary types and forbids extra fields (saves us from silent failures due to typos for optional fields)

### DIFF
--- a/lcls_tools/common/__init__.py
+++ b/lcls_tools/common/__init__.py
@@ -1,0 +1,3 @@
+from lcls_tools.common.pydantic import BaseModel
+
+__all__ = ["BaseModel"]
--- a/lcls_tools/common/devices/area.py
+++ b/lcls_tools/common/devices/area.py
@@ -20,10 +20,12 @@ from lcls_tools.common.devices.wire import (
     WireCollection,
 )
 
-from pydantic import BaseModel, SerializeAsAny, Field, field_validator
+from pydantic import SerializeAsAny, Field, field_validator
+
+from lcls_tools.common.pydantic import LCLSToolsBaseModel
 
 
-class Area(BaseModel):
+class Area(LCLSToolsBaseModel):
     """This class provides access to collections of hardware components
     in a given machine area of LCLS/LCLS-II (for example: BC1, or BC2).
     The information for each collection is provided in YAML configuration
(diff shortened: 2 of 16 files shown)


robnagler (approved):
Seems like a good idea. I do question the need for the complex name. Maybe even just lcls_tools.BaseModel, since it's simple enough, and it's needed everywhere.

roussel-ryan:
I guess we would want something different to not accidentally use the pydantic BaseModel somewhere, maybe something simpler? 'LCLSBaseModel`?

eloiseyang:
lume-model does something similar and calls its base model LUMEBaseModel. I think either name is fine, but LCLSBaseModel is consistent with lume-model, another SLAC pydantic code.

roussel-ryan:
LCLSBaseModel sounds good to me then

robnagler:
> I guess we would want something different to not accidentally use the pydantic BaseModel somewhere, maybe something simpler? 'LCLSBaseModel`?

This is checkable, if need be. There's nothing preventing someone from using `pydantic.BaseModel` so I don't really see what protection is being added.  I don't see the need for a check either.

I will stress (yet again) that the style of using unqualified imports (especially in this case) is less robust than making always use a qualified import. This has many advantages ([our view](https://github.com/radiasoft/pykern/wiki/DesignHints#qualify-imports) and 

Perhaps consider `lclstools.Base`, `lclstools.Model` or `lclstools.Container`. This is a central class in the library, used by most classes in the library. One could argue it should be used by all classes if you are buying into Pydantic as a typing system. I'm not arguing this or the use of Pydantic.

robnagler:
FWIW, ChatGPT answers "qualified imports in python vs not" with:

> When to Use What?
> - Use qualified imports (import module) when working with large projects or when name conflicts are possible.
> - Use unqualified imports (from module import X) when using a small number of specific functions or classes frequently.
> - Use aliases (import module as alias) when the module name is long but you still want clarity.

I agree with this.

roussel-ryan:
All right, that seems good

roussel-ryan:
In that case, should we have the pydantic class defined in lcls_tools instead of lcls_tools.common? This also begs the question of the existence of the common directory @eloiseyang @nneveu thoughts?

roussel-ryan:
> In that case, should we have the pydantic class defined in lcls_tools instead of lcls_tools.common? This also begs the question of the existence of the common directory @eloiseyang @nneveu thoughts?

So for now I put it into lcls_tools.common to be consistent with everything else, we can do a big refactor at a later time if necessary. I've also qualified all of the imports for the BaseModel based on @robnagler 's suggestion and fixed some tests that failed due to the stricter field checking

PR #238, opened 2025-02-21
Adding refresh rate pvs and property to screens.

phys-cgarnier:
Hi, this pull request is to add refresh rate PVs and a getter method to the screens device class.
Some info that needs to be known beforehand is that many cameras have two refresh rate PVs, but only one is used. In order to determine what PV to use you need to know whether the IOC the camera runs on is a VME or LinuxRT. So I added the PV that tells you that information as well.


roussel-ryan:
can you add a set of tests to the device testing file that validates the property values

roussel-ryan:
closed because the PR comes from a fork and breaks github install actions

PR #239, opened 2025-02-21
Add linting

pluflou:
- added ruff formatting and pre-commit hooks to enforce one style across the codebase
- dropped flake8 in favor of ruff
- cleaned up codebase to clear formatting/linting errors based on ruff configs (mostly defaults, can adjust based on preference)
- added GHA workflow for running pre-commit hook on all files
- updated README to reflect this change


docs/source/conf.py
```diff
@@ -6,14 +6,14 @@
...
-from lcls_tools.common.data.fitting_tool import *
-from lcls_tools.common.devices.magnet import *
-from lcls_tools.common.devices.beampath import *
-from lcls_tools.common.devices.device import *
-from lcls_tools.common.devices.screen import *
-from lcls_tools.common.matlab2py.mat_image import *
-from lcls_tools.common.matlab2py.mat_emit_scan import *
-from lcls_tools.common.matlab2py.mat_corr_plot import *
+from lcls_tools.common.data.fitting_tool import *  # noqa: F403
+from lcls_tools.common.devices.magnet import *  # noqa: F403
```
pluflou, line 10: I don't think these imports are needed in a conf.py, only the necessary paths for the build. But I could be wrong. Can anyone comment?

lcls_tools/common/image/__init__.py
```diff
@@ -1 +1 @@
-from lcls_tools.common.image.image import Image
+from lcls_tools.common.image.image import Image as Image
```
eloiseyang, line 1: I imagine this is just the linter working, but is there an advantage to this?

eloiseyang (approved):
Hi Sara, thanks for the PR. It's nice to have a linter that makes the changes directly. I don't have experience with Sphinx, so I can't answer your question. Let's get that resolved before merging. I'm going to approve anyhow.

lcls_tools/common/image/__init__.py
pluflou: The linter doesn't understand unused imports in an `__init__` file. Importing it as an alias is one way to solve that. The other possibly preferred options are to tell it to ignore these violations in all `__init__` files, or using `__all__`:
```
from lcls_tools.common.image.image import Image

__all__ = ["Image"]
```
See more info [here](https://docs.astral.sh/ruff/rules/unused-import/). Let me know if you have a preference on how to handle these.

pluflou:
> I don't have experience with Sphinx, so I can't answer your question. Let's get that resolved before merging. I'm going to approve anyhow.

From what I can tell running the docs locally and from my experience setting up sphinx sites, lines 8-16 are not needed. But I was hoping to confirm with the author of the conf.py file in case this was intentional.

lcls_tools/common/image/__init__.py
eloiseyang: The redundant alias is fine to me, I just didn't understand it. Thanks for the explanation.

robnagler: Seems to me would be better to either define Image in common.image or not have the alias. Aliases are problematic.

docs/source/conf.py
robnagler: Our [conf.py](https://github.com/radiasoft/pykern/blob/79511e45c374c25d6971cef834a316769bd29427/pykern/package_data/sphinx-conf.py) has no imports. We [dynamically generate it](https://github.com/radiasoft/pykern/blob/79511e45c374c25d6971cef834a316769bd29427/pykern/pkcli/sphinx.py#L22) before sphinx runs so it is package independent and doesn't have to be maintained. There are only four package specific variables:

```py
# Dynamically filled variables
author = """$author"""
project = """$name"""
version = """$version"""
year = """$year"""
```

And we don't use jinja here because it doesn't work well with files that have a lot of f-strings.

lcls_tools/common/image/__init__.py
nneveu: I'm not a fan of this import structure either. This file is a hold over and should have been removed when fit/processing was merged. There are now 2 files that I'm aware of that should have been removed w/ the image PR merges. I'll start a new PR to remove those. Then we shouldn't need this import.

roussel-ryan: Perfect, I've been looking for a reason to remove this file

roussel-ryan: I wasn't sure if it was being used by someone else.

docs/source/conf.py
nneveu: IIRC were added to get the indexing on the page, but if there's an automatic way to do it, let's go ahead and get rid of these.

lcls_tools/common/image/__init__.py
eloiseyang: Let me see if I understand @robnagler. This line essentially bumps the `Image` class up by one subpackage, such that the user only has to write:
```
from lcls_tools.common.image import Image
```
Instead of:
```
from lcls_tools.common.image.image import Image
```
The issue with this solution is that the Image class is now an alias on top of an unqualified import, which can cause namespacing issues like you describe in [this article](https://www.robnagler.com/2014/06/27/PythonTip-Import-From-Considered-Harmful.html). The solution in the future would be to use a qualified import and avoid aliasing. To do this, we would move the Image class up subpackage, on the level of `lcls_tools.common.image`. Then the following qualified import would be the correct use:
```
import lcls_tools.common.image
...
data = lcls_tools.common.image.Image(args)
```
Even though this class call is bulky, it's more Python safe. The solution to the long class call would probably be to break up `image` into a separate package.

Is this more or less what you're saying?

robnagler: Correct, and I would remove `common` from the hierarchy. I realize this breaks code, but for that, you can put in a backward compatible stub with a deprecation warning.

eloiseyang:
Merged now. Thanks @pluflou!

PR #240, opened 2025-02-21
Refresh rate

phys-cgarnier:
Hi, this pull request is to add refresh rate PVs and a getter method to the screens device class.
Some info that needs to be known beforehand is that many cameras have two refresh rate PVs, but only one is used. In order to determine what PV to use you need to know whether the IOC the camera runs on is a VME or LinuxRT. So I added the PV that tells you that information as well.


phys-cgarnier:
from @roussel-ryan:
 "can you add a set of tests to the device testing file that validates the property values"

phys-cgarnier:
still setting up unit test will complete soon.

eloiseyang:
Hi @phys-cgarnier. Looks like the linter is still unhappy. Could you fix the flake8 problems so that the unit tests can run? I can help you out after that.

eloiseyang (approved):
Hey Ryan, thanks for the PR. Looks like the only failing tests are the SC ones, so you can go ahead and ignore those. Just merge with main and I'll be ready to approve. We should let the new linter make a once over. The changes themselves look good, so I'll go ahead and approve.

nneveu:
@phys-cgarnier Eloise approved this, so this can merge after doing linter and conflict fixes.

PR #244, opened 2025-02-24
Specify shell for running tests

pluflou:
This addresses a bug in the testing workflow where tests fail but the GHA show as passed, see here:
- [example 1](https://github.com/slaclab/lcls-tools/actions/runs/13459888949/job/37612415062)
- [example 2](https://github.com/slaclab/lcls-tools/actions/runs/13478922929/job/37661632846)

This is due to how the GitHub step summary is set up to print the output, where the “>>” operator doesn’t play nice with default shells. This PR sets the shell explicitly and resolves this issue.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,6 +28,7 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
+      shell: bash
       run: |
         eval `ssh-agent -s`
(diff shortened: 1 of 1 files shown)


nneveu:
Can you add a description of what issue exactly this PR is addressing, and how the problem is solved?

pluflou:
> Can you add a description of what issue exactly this PR is addressing, and how the problem is solved?

Done (this was based on Ryan's reports of the issue on the lcls-tools slack channel).

roussel-ryan (approved):

PR #249, opened 2025-03-06
Removed all the secrets related code and the code that installs meme.

eloiseyang:
I removed all the stuff that uses git secrets so we can delete the secret. Once this is merged the actions can run on tests from forks again.


nneveu (approved):
I'm approving this, but I think there is one more change needed. We'll need to add the import where meme is called, or remove that meme call completely? It doesn't look like it is even needed here, but I could be missing something: https://github.com/slaclab/lcls-tools/blob/9151bc4f4929259aa6f956cec22a0b1c167d7c72/tests/unit_tests/lcls_tools/common/measurements/test_emittance_measurement.py#L40

 Regardless, this can be merged once it passes the new linting. Thanks!

eloiseyang:
Hi @nneveu, I think you're right, this PR is kind of a WIP. I started it with the plan to finish it right away, then I got some other requests...

eloiseyang:
@nneveu It was needed, once upon a time. Let's cut that part of the emittance measurement test for now. The correct thing to do would be to mock it out.

PR #251, opened 2025-03-12
More tools for dual energy matching

pluflou:
Creating this for @williamColocho 
Adding tools for energy gain in dual energy matching mode. The Dump energy PV needs to be determined dynamically depending on primary/secondary beam path.


lcls_tools/common/data/bmad_modeling/bmad_modeling.py
```diff
@@ -5,11 +5,12 @@
 from datetime import datetime, timedelta
 from lcls_live.datamaps import get_datamaps
 from lcls_live.archiver import lcls_archiver_restore
-
+from lcls_tools.common.data.model_general_calcs import bdes_to_kmod
 YAML_LOCATION = os.path.join(os.path.dirname(__file__), "yaml/")
+print(YAML_LOCATION)
```
eloiseyang, line 10: What is this print statement trying to accomplish? The only time this print statement will run is when we import the `bmad_modeling` path, which is not intuitive for the user.

lcls_tools/common/data/bmad_modeling/bmad_modeling.py
```diff
@@ -45,6 +47,7 @@ def get_energy_gain_pvlist(beam_path):
     except FileNotFoundError:
         print(f"Could not find yaml file {yaml_file_name}")
```
eloiseyang, line 48: I'd like to throw an error here. It can provide more information than the regular FileNotFoundError does. I don't think there's a situation where the caller of this function wants to proceed even if there's a file not found error. If that's the case, the caller can wrap this function in their own try/catch statement.

eloiseyang (commented):
Hi William, thanks for the PR. I left a couple of comments. Once we get those resolved, the PR will be ready to merge. Please merge with main so that the tests run as they should. We can move this PR back to your fork to make it easier for you to change.

eloise-nebula:
@williamColocho any updates on this PR? If it's on hold can we close it for now?

nneveu:
@pluflou is this still a work in progress? I know William has been busy w/ beam times.

pluflou:
@nneveu I'm not sure. I just created the PR because he didn't have the permissions at the time. I am not working on the actual stuff in the PR.

PR #252, opened 2025-03-12
Add basic BPM device

kabanaty:
This code implements a basic BPM device with only one BPM, along with the necessary code to update YAML files and create the object.  
This implementation only uses one BPM PV, being :TMIT.  This is to jump start development of TMIT Loss measurements using lcls-tools.


lcls_tools/common/devices/bpm.py
```diff
@@ -0,0 +1,108 @@
...
+        super().__init__(**kwargs)
+
+    @field_validator("*", mode="before")
+    def validate_pv_fields(cls, v: str) -> PV:
+        return PV(v)
+
+
+class BPMControlInformation(ControlInformation):
+    PVs: SerializeAsAny[BPMPVSet]
+    _ctrl_options: SerializeAsAny[Optional[Dict[str, int]]] = dict()
```
eloiseyang, line 52: Looks like this was copied over from magnet.py. Can we remove this since it isn't initialized for now? Maybe just comment it out with a TODO.

lcls_tools/common/devices/bpm.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super(BPMControlInformation, self).__init__(*args, **kwargs)
+
+
+class BPMMetadata(Metadata):
```
eloiseyang, line 58: Could you add a TODO comment here? Since it's empty right now.

lcls_tools/common/devices/bpm.py
```diff
@@ -0,0 +1,108 @@
...
+    def validate_bpms(cls, v) -> Dict[str, BPM]:
+        for name, bpm in v.items():
+            bpm = dict(bpm)
+            # Set name field for BPM
+            bpm.update({"name": name})
+            v.update({name: bpm})
+        return v
+
+    # TODO: can the next two functions get moved out?
+    def seconds_since(self, time_to_check: datetime) -> int:
```
eloiseyang, line 94: What's this function for? It's not called in this PR as far as I can see.

lcls_tools/common/devices/bpm.py
```diff
@@ -0,0 +1,108 @@
...
+        if not isinstance(time_to_check, datetime):
+            raise TypeError("Please provide a datetime object for comparison.")
+        return (datetime.now() - time_to_check).seconds
+
+    def _make_bpm_names_list_from_args(
+        self, args: Union[str, List[str], None]
+    ) -> List[str]:
+        bpm_names = args
+        if bpm_names:
```
eloiseyang, line 103: Nested if statements can be difficult to parse. This one isn't difficult to read, but I would probably write this one as:

```
if ~bpm_names:
    bpm_names = list(self.bpms.keys())
    return bpm_names
if isinstance(bpm_names, str):
    bpm_names = [args]
else:
    bpm_names = args
return bpm_names
```

lcls_tools/common/devices/bpm.py
```diff
@@ -0,0 +1,108 @@
...
+            if isinstance(bpm_names, str):
```
eloiseyang, line 104: I realized while writing the above comment that you forgot to handle the case where args is a list of strings. Whether or not you use my solution you should cover that case.

lcls_tools/common/devices/reader.py
```diff
@@ -142,6 +143,24 @@ def create_lblm(
...
 
+def create_bpm(
+    area: str = None, name: str = None
+) -> Union[None, BPM, BPMCollection]:
+    device_data = _device_data(area=area, device_type="bpms", name=name)
+    if not device_data:
+        return None
+    if name:
+        try:
+            # this data is not available from YAML directly in this form, so we add it here.
```
eloiseyang, line 154: Is there a way to grab this information programmatically? If the answer is no, that's okay for this first pass. It does kind of defeat the purpose of YAMLs if we have to add the name manually every time though. I can take a crack at it later too.

lcls_tools/common/devices/reader.py
```diff
@@ -142,6 +143,24 @@ def create_lblm(
         return LBLMCollection(**device_data)
 
 
+def create_bpm(
+    area: str = None, name: str = None
+) -> Union[None, BPM, BPMCollection]:
+    device_data = _device_data(area=area, device_type="bpms", name=name)
+    if not device_data:
+        return None
```
eloiseyang, line 151: I don't love this pattern in general, though it's not your issue. I feel like `_device_data` should throw an error if it fails to find a YAML file. It leads to situations like this where the caller, `create_bpm`, has to catch a silent failure, then communicate that silent failure to its own caller by returning `None`. The same effect could be achieved if `_device_data` would just throw its own error.

I know there are situations where the caller might want to ignore the fact that a BPM can't be initialized, as in the case of mass BPM initialization. In that case, the caller can just create their own try/catch block.

That being said, adding errors to `_device_data` is probably out of scope for this PR. If people agree with me, I can write an issue and self assign it.

lcls_tools/common/devices/reader.py
```diff
@@ -142,6 +143,24 @@ def create_lblm(
...
+    if name:
+        try:
+            # this data is not available from YAML directly in this form, so we add it here.
+            device_data.update({"name": name})
+            return BPM(**device_data)
+        except ValidationError as field_error:
```
eloiseyang, line 157: In a similar vain to the above comment, I think we should just pass the exception here, maybe with a bespoke error message. If the caller needs to ignore the error, they can do so with their own try/catch block.

eloiseyang (commented):
Hi Tyler, thanks for the PR. I left a few comments, so let me know your thoughts. A couple requests: Could you write in the description that this PR only implements the TMIT PV? Could you also create a simple test for the bpm like in wire? Just for the TMIT PV if that's possible.

Looks like the linter is mad too. Sara provided the following instructions on Slack to run the linter locally:
> Install pre-commit with `pip install pre-commit` and then run `pre-commit install` in the root of the repository, then run `pre-commit run --all-files` to check all files

Thanks again! Let me know if you have any questions or need help.

lcls_tools/common/devices/bpm.py
kabanaty: BPMControlInformation is needed to initialize the PVs that are read out as properties (single PV in this case)

eloiseyang: I see now, thank you for your explanation.

lcls_tools/common/devices/bpm.py
kabanaty: I think this is a holdover from other device classes... not sure it's used anywhere but I removed it.

lcls_tools/common/devices/bpm.py
kabanaty: I think I will leave the code as-is for consistency's sake with the other device files. I believe this code already handles args as a list of strings.  If `args = List[str]`, when `bpm_names = args` is evaluated, `bpm_names `will be output as `List[str]` as demanded by the validator.

lcls_tools/common/devices/bpm.py
kabanaty: I agree that this is more readable, but for consistency's sake I will leave the code as is.

lcls_tools/common/devices/reader.py
kabanaty: `name` is used to instantiate the object, e.g. `my_bpm = create_bpm(area="BYP", name="BPMBP15").  I'm not sure that there is a way to programmatically get the name, because how else would the function know which device-object to make? I also may not be understanding your question.

lcls_tools/common/devices/bpm.py
eloiseyang: Oh, you're right. Sounds good.

lcls_tools/common/devices/reader.py
eloiseyang: You're right, I thought name was being initialized as an attribute somehow.

lcls_tools/common/devices/reader.py
kabanaty: I'm in support of the proposed `_device_data` fix

lcls_tools/common/devices/reader.py
kabanaty: Also in support of changing exception-passing to something more intuitive.

eloiseyang (approved):
Thanks for replying to my changes! I can write tests when I take over this device.

PR #256, opened 2025-03-26
Add X and Y PVs to BPM Device and YAMLs

eloiseyang:
Tyler wrote a PR to create the BPM devices with just the TMIT he needed, and I agreed to finish it out.

I double checked that the BPMs didn't have relevant metadata in the Oracle database, and I couldn't find anything to add. When you query the BPMs for attribute PVs, you can find somewhere between 90-2000 attributes. I looked at some of the relevant GUIs, and it looks like only X, Y, and TMIT are used regularly. After some more discussion with other folks, I decided it was right just to include X and Y.

So here's the final product, a pretty basic PR that just adds X and Y PVs for BPMs.


kabanaty:
Hi Eloise, have you added methods for getting these PVs? Or does that come for free since we don't need a setter?

eloiseyang:
Hi @kabanaty, good call I missed a couple spots. I added the buffer information too, and I changed the TMIT buffer to something I thought makes more sense. Let me know if there was a specific reason why it was the way it was before though.

kabanaty:
These changes look great! The docstrings need to be updated, they're still the copy/paste from TMIT though.

lcls_tools/common/devices/yaml/LI24.yaml
nneveu, line 1: Do we want this area included?

nneveu (approved):
Looks good! I left a few comments, but they don't need to prevent merging. You might need to update w/ main, but otherwise merge when ready.

lcls_tools/common/devices/yaml/LI24.yaml
eloiseyang: Not really! I will remove it.

lcls_tools/common/devices/reader.py
```diff
@@ -145,6 +145,7 @@ def create_lblm(
 
 def create_bpm(area: str = None, name: str = None) -> Union[None, BPM, BPMCollection]:
     device_data = _device_data(area=area, device_type="bpms", name=name)
+    print(device_data)
```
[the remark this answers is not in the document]
eloiseyang: That was for testing, I'll remove it.

kabanaty (approved):

PR #259, opened 2025-04-09
Util Gauss Fit Import

eloise-nebula:
This PR adds the profiles available in our Matlab Gaussian fitting code, util_gaussFit, including Gaussian, asymmetrical Gaussian, super Gaussian, and asymmetrical super Gaussian profiles. The following links contain definitions for the asymmetrical Gaussian and super Gaussian respectively, and the asymmetrical super Gaussian is a combination of the two.

(The first equation in section V. Method of Data Analysis)
https://journals.aps.org/prab/pdf/10.1103/PhysRevSTAB.3.034401
https://www.rp-photonics.com/flat_top_beams.html

Henrik includes the following term for the skew of the asymmetric Gaussian:
```
% Line 253 of util_gaussFit
E = sign(A) * min(1 - 1e-10, abs(A));
```
Where A is a term controlling the proportion of skew. This code functionally limits the value of A to be (-1, 1) exclusive, so the correct approach to applying this in the new code is to use scipy's bounds options. Each profile has an associated wrapper for the `scipy.optimize.curve_fit` function with appropriate bounds.

A few notes on implementation I'd like feedback on:

The `__fit__` function creates a list of parameters using the argument names of the input curves. While this is cute, and saves me from writing about 5 lines of code, it may not be obvious or robust.

In the Bayesian implementation of this same idea, I requested that we include some parameter classes to wrap the output of our fit and image statistics functions. I requested that because I remembered how distressed I was when our Matlab code would provide a struct as function input, and get back a struct with changed values and new properties as output. Thinking about it now, I think using dictionaries is fine, so long as we make it obvious what the keys are in the function that outputs it. e.g. profile_stat and image_stat use the literal definition to create its output dictionaries.

That being said, it would be nice to create an image class that can makes basic image manipulation easy. For instance:
```
image_1 = lcls_tools.profmon.image(data_1)
image_2 = lcls_tools.profmon.image(data_2)
background = lcls_tools.profmon.image(bg)
image_1 = image_1 - background
image_2 = image_2 - background
mean_image = lcls_tools.profmon.mean([image_1, image_2])
```

Those are my next plans, but I thought a PR with that class and this stuff would be too big.


lcls_tools/common/data/least_squares.py
```diff
@@ -0,0 +1,42 @@
...
+import lcls_tools.common.data.profile
+import numpy
+import scipy.optimize
+
+
+def __fit__(curve, x, y, **kwargs):
+    popt, cov = scipy.optimize.curve_fit(curve, x, y, **kwargs)
+    sig = inspect.signature(curve)
+    # Grab params excluding x.
+    par = list(sig.parameters)[1:]
```
nneveu, line 11: I agree on your comment in the PR write up. I'd prefer to do this w/o a hard coded 1, but leaving it for now. It will be easy enough to change later.

lcls_tools/common/data/profile.py
```diff
@@ -0,0 +1,27 @@
+import math
+import numpy
+
+
+def gaussian(x, mu, sigma):
```
nneveu, line 5: Why not use scipy.stats.norm and scale? Or is this to exactly match Matlab?

lcls_tools/common/image/profmon.py
```diff
@@ -0,0 +1,32 @@
...
+curves = {
+    "gaussian": lcls_tools.common.data.least_squares.gaussian,
+    "asymmetrical_gaussian": lcls_tools.common.data.least_squares.asymmetrical_gaussian,
+    "super_gaussian": lcls_tools.common.data.least_squares.super_gaussian,
+    "asymmetrical_super_gaussian": lcls_tools.common.data.least_squares.asymmetrical_super_gaussian,
+}
+
+
+def profile_stat(profile, curve="gaussian"):
+    fit = curve[curve]
```
nneveu, line 13: should this be `curves[curve]`? missing 's'?

lcls_tools/common/image/profmon.py
```diff
@@ -0,0 +1,32 @@
...
+    y_stat = profile_stat(y_projection, curve=curve)
+    centroid = [x_stat["centroid"], y_stat["centroid"]]
+    rms_size = [x_stat["centroid"], y_stat["centroid"]]
+    stat = {
+        "centroid": centroid,
+        "rms_size": rms_size,
+        "intensity": image.sum(),
+        "curve": curve,
+    }
+    return stat
```
nneveu, line 32: We might want to return more info here? Or maybe I'm missing something. i.e. if someone wants to plot the projection w/ the stat, we'd need to return the profiles too?

nneveu (approved):
Thanks for this great PR! I left some simple comments, but looks good.

shamin-slac (approved):
Looks good to me (just need to fix a few typos). 

I think separating the fitting from the specific functions for the different curves is good, and it makes it easy to add new functions.

Returning dictionaries is fine, especially since it's clear what they are based on the names and you've shown what's in them (though it wouldn't hurt to have some function docstrings)

I think the image class is a good idea.

lcls_tools/common/data/least_squares.py
```diff
@@ -0,0 +1,42 @@
...
+    return params, cov
+
+
+def gaussian(x, y):
+    gauss = lcls_tools.common.data.profile.gaussian
+    return __fit__(gauss, x, y)
+
+
+def asymmetrical_gaussian(x, y):
+    gauss = lcls_tools.common.data.profile.asymmetrical_gaussian
```
kabanaty, line 24: It's a little confusing that all the fit methods define a variable called "gauss".  Maybe "asym_gauss", "super_gauss", etc.?

kabanaty:
I'm late to the party but everything looks good to me! I agree with the points made by Nicole and Shamin.  Definitely ready to merge but a couple of small changes to clean it up and get it in great shape.

kabanaty (approved):

lcls_tools/common/data/least_squares.py
```diff
@@ -0,0 +1,42 @@
+import inspect
+import lcls_tools.common.data.profile
+import numpy
+import scipy.optimize
+
+
+def __fit__(curve, x, y, **kwargs):
```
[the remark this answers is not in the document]
kabanaty: __fit__ implies a native Python method.  "__fit" is a good choice as it's 'private', but I think "fit" would be sufficient as well.

lcls_tools/common/data/least_squares.py
kabanaty: The 5 extra lines would be worth the clarity here.

lcls_tools/common/data/profile.py
eloiseyang: This function doesn't generate a random distribution of Gaussian points like scipy.stats.norm does. The functions in profile.py are the Gaussian-family profiles that we use to fit our diagnostics. The scipy curve fitting function, scipy.optimize.curve_fit, takes a function of the form f(x, param_1, param_2, ...) as input, and outputs a list of the form [fitted_param_1, fitted_param_2, ...]. These profiles define f and give an easy way to plot the resulting curve after.

lcls_tools/common/image/profmon.py
eloiseyang: It should be! I might change the names here so that this kind of thing is easier to catch.

lcls_tools/common/image/profmon.py
eloiseyang: I agree that we will probably iterate on this function to return more stats, but I'd rather not package the profile with the stats output. It should be the caller's responsibility to know what profile it used, and it can package the profile with the stats for someone who might not know what profile was used, say if the data were saved somewhere.

lcls_tools/common/data/least_squares.py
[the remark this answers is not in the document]
eloiseyang: Makes sense, I just did what "looked right" to me. I'll use the appropriate Python syntax.

lcls_tools/common/image/profmon.py
nneveu: maybe `curve_list` or `curves_list`

lcls_tools/common/image/profmon.py
nneveu: So the user would have to also call the correct profile function outside of this and know that the corresponding stats match? i.e. would it be easy for someone to accidentally calc stat for a gaussian, but plot a super gaussian profile? Maybe the "curve" key pair is enough so that they should be able to avoid this, like you said. I agree let's leave it for now, but maybe put in some checks when people plot profiles/stats together.

nneveu (approved):

lcls_tools/common/data/profile.py
eloiseyang: Sorry, I just saw that you can use scipy.stats.norm.pdf. We could do that, but I like having the equation written explicitly in this case. If you'd rather use the scipy pdf we could do that.

lcls_tools/common/image/profmon.py
eloiseyang: I guess I don't see the situation where the user doesn't know what profile they used. The typical use cases will look something like this:
```
import lcls_tools.common.image.profmon as profmon

# default case is labeled in function definition & can be documented
stat = profmon.image_stat(image)
fit = profmon.curve_dict["gaussian"]
plot = fit(x, stat["centroid"], stat["rms_size"])


profile = "super_gaussian"
profmon.image_stat(image, curve=profile)
fit = profmon.curve_dict[profile]
plot = fit(x, stat["centroid"], stat["rms_size"], stat["other"])
```
It's true that someone could be ignorant of the default case or ignore what the code they wrote says, but they could just as easily ignore another dictionary value that tries to nudge them towards the correct profile. Adding the same information in another place doesn't make the functionality of the code clearer imo.

nneveu (approved):

PR #260, opened 2025-04-16
Wire Scan Measurement

kabanaty:
This pull request adds code for doing a Wire Scan Beam Profile Measurement.  In addition, this PR will introduce code for calculating TMIT Loss during these measurements.  There are also small fixes to `bpm`, `lblm`, and `wire `device code that were discovered during development of the measurement.  

Note that in `wire_scan` there are some direct-to-epics calls that I would like to phase out in the future.  Currently, the destination controls in the `edef` module are pointing at the wrong PVs and are effectively useless, per R. McClanahan of the timing team.  I hope to submit a PR to `edef` or have M. Gibbs work on a fix soon.


lcls_tools/common/measurements/wire_scan.py
```diff
@@ -0,0 +1,335 @@
...
+import os
+from pydantic import ConfigDict, SerializeAsAny, BaseModel
+from lcls_tools.common.measurements.utils import NDArrayAnnotatedType
+from lcls_tools.common.measurements.tmit_loss import TMITLoss
+import numpy as np
+import pandas as pd
+import epics
+
+
+class WireBeamProfileMeasurementResult(BaseModel):
```
shamin-slac, line 17: Can we include rms_sizes in the result? The emittance measurement class needs them to calculate emittance and bmag

lcls_tools/common/measurements/tmit_loss.py
```diff
@@ -0,0 +1,358 @@
...
+                          - "HTR", "DIAG0", "COL1", "EMIT2", "BYP",
+                            "SPD", "LTUS".
+            bpms_devices (list): A list of BPM device names.
+
+        Returns:
+            tuple: A tuple containing:
+                - list: Indices of BPMs located **before** the wire.
+                - list: Indices of BPMs located **after** the wire.
+        """
+        # Define valid regions
```
shamin-slac, line 183: It would be better to have these region and bpm names identified externally (maybe using the yaml files)

shamin-slac (commented):
Overall I think it's fine. In particular, the wire scan measurement result needs the rms_sizes, so the emittance measurement can use it. Also, it would be good to have test(s) for these measurements.

lcls_tools/common/measurements/tmit_loss.py
kabanaty: I'm open to this idea but I'm not sure how to implement it.  Adding TMITLOSS metadata seems like the obvious way but we would have to manage the metadata on a per-bpm basis instead of a per-area basis.  Until I have time to think about this more I'd like to keep the current implementation.

kabanaty:
Closing pull request for the time being to do more testing.

PR #261, opened 2025-04-18
Create a new device class for tcavs

phys-cgarnier:
TCAV Device class:
Changed generator.py to include extract_tcav function. Also implemented advanced filtering in extract devices. You can pass an unpacked dictionary that is organized by  {field_name: expected value}
For example :
additional_filter_constraints = {"Engineering Name": "TRANS_DEFL"}
Added relevant fields to metadata.py


lcls_tools/common/devices/tcav.py
```diff
@@ -0,0 +1,228 @@
...
+    PVs: SerializeAsAny[TCAVPVSet]
+    _mode_config_options: SerializeAsAny[Optional[Dict[str, int]]] = dict()
+
+    def __init__(self, *args, **kwargs):
+        super(TCAVControlInformation, self).__init__(*args, **kwargs)
+
+    def set_mode_config_option(self):
+        mode_config_options = self.PVs.mode_configs.get_ctrlvars(timeout=1)
+        if mode_config_options:
+            [
```
eloise-nebula, line 52: I think it's fewer lines to write this for loop as:
```
for i, option in enumerate(mode_config_options["enum_strs"]):
    self._mode_config_options.update({option: i})
```

lcls_tools/common/devices/tcav.py
```diff
@@ -0,0 +1,228 @@
...
+
+class TCAVControlInformation(ControlInformation):
+    PVs: SerializeAsAny[TCAVPVSet]
+    _mode_config_options: SerializeAsAny[Optional[Dict[str, int]]] = dict()
+
+    def __init__(self, *args, **kwargs):
+        super(TCAVControlInformation, self).__init__(*args, **kwargs)
+
+    def set_mode_config_option(self):
+        mode_config_options = self.PVs.mode_configs.get_ctrlvars(timeout=1)
```
eloise-nebula, line 50: Should this code error out if `get_ctrlvars` times out? What's the use case for a TCAV without `mode_config_options`?

lcls_tools/common/devices/tcav.py
```diff
@@ -0,0 +1,228 @@
...
+        """Decorator to only allow :MODECFG to be set if that option exists for the TCAV"""
+
+        def decorator(function):
+            @wraps(function)
+            def decorated(self, *args, **kwargs):
+                if isinstance(options_to_check, str):
+                    options = [options_to_check]
+                for option in options:
+                    if option not in self.controls_information.mode_config_options:
+                        print(
```
eloise-nebula, line 87: I think an error here would be more appropriate. As it stands, the `check_options` decorator doesn't do anything to communicate to the code that something went wrong.

lcls_tools/common/devices/tcav.py
```diff
@@ -0,0 +1,228 @@
...
+    def check_state(f):
+        """
+        Decorator to enforce that the device is not in 'Disable'
+        mode before executing an operation. Prevents execution of the
+        decorated method if `mode_config == "Disable"`.
+        """
+
+        def decorated(self, *args, **kwargs):
+            if self.mode_config == "Disable":
+                print("Unable to perform action, TCAV is in Disabled state")
```
eloise-nebula, line 106: Another place where an error might be better than a print statement.

lcls_tools/common/devices/tcav.py
```diff
@@ -0,0 +1,228 @@
...
+        return decorated
+
+    @property
+    def amp_set(self):
+        """The amplitude set point of the TCAV"""
+        return self.controls_information.PVs.amp_set.get()
+
+    @amp_set.setter
+    @check_state
+    def amp_set(self, amplitude):
```
eloise-nebula, line 119: Pydantic lets you validate function arguments with function annotations like so:
```
def amp_set(self, amplitude: float):
```
The if statement shouldn't be necessary.

lcls_tools/common/devices/tcav.py
```diff
@@ -0,0 +1,228 @@
...
+        self.controls_information.PVs.amp_set.put(amplitude)
+
+    @property
+    def phase_set(self):
+        """The phase set point of the TCAV"""
+        return self.controls_information.PVs.phase_set.get()
+
+    @phase_set.setter
+    @check_state
+    def phase_set(self, phase):
```
eloise-nebula, line 131: Another place for pydantic validation.

lcls_tools/common/devices/tcav.py
```diff
@@ -0,0 +1,228 @@
...
+
+    @property
+    def amp_fbenb(self):
+        """The status of the amplitude set point feedback"""
+        return self.controls_information.PVs.amp_fbenb.get()
+
+    @amp_fbenb.setter
+    @check_state
+    def amp_fbenb(self, state: Union[str, int]):
+        if not isinstance(state, str) or not isinstance(state, int):
```
eloise-nebula, line 144: Pydantic should already be validating this.

lcls_tools/common/devices/tcav.py
```diff
@@ -0,0 +1,228 @@
...
+
+    @property
+    def phase_fbenb(self):
+        """The status of the phase set point feedback"""
+        return self.controls_information.PVs.phase_fbenb.get()
+
+    @phase_fbenb.setter
+    @check_state
+    def phase_fbenb(self, state: Union[str, int]):
+        if not isinstance(state, str) or not isinstance(state, int):
```
eloise-nebula, line 156: Another place where pydantic covers function validation.

lcls_tools/common/devices/tcav.py
```diff
@@ -0,0 +1,228 @@
...
+
+    @property
+    def amp_fbst(self):
+        """The state of the amplitude feedback"""
+        return self.controls_information.PVs.amp_fbst.get()
+
+    @amp_fbst.setter
+    @check_state
+    def amp_fbst(self, state: Union[str, int]):
+        if not isinstance(state, str) or not isinstance(state, int):
```
eloise-nebula, line 168: Same as above.

lcls_tools/common/devices/tcav.py
```diff
@@ -0,0 +1,228 @@
...
+
+    @property
+    def phase_fbst(self):
+        """The state of the phase feedback"""
+        return self.controls_information.PVs.phase_fbst.get()
+
+    @phase_fbst.setter
+    @check_state
+    def phase_fbst(self, state: Union[str, int]):
+        if not isinstance(state, str) or not isinstance(state, int):
```
eloise-nebula, line 180: Same as above.

lcls_tools/common/devices/tcav.py
```diff
@@ -0,0 +1,228 @@
...
+        self.controls_information.PVs.mode_config.put("DISABLE")
+
+    @property
+    def l_eff(self):
+        """The effective length of the TCAV in meters"""
+        return self.metadata.l_eff
+
+    @l_eff.setter
+    def l_eff(self, length):
+        if not isinstance(length, float):
```
eloise-nebula, line 208: Pydantic validation could replace this.

eloise-nebula (commented):
Hi Chris, looks consistent with our other device classes. I found a bunch of places where pydantic validation could replace your manual validation. This device is unique because its configuration depends on its mode_config, but I think that's handled well here. Once we figure out the validation stuff.

lcls_tools/common/devices/tcav.py
phys-cgarnier: Sure I can do that, I was trying to make this as close to a match to magnet.py as possible but I can clean it up.

lcls_tools/common/devices/tcav.py
phys-cgarnier: I believe (although am unable to confirm at this time), that you cannot or should not be setting other pvs in the tcav while TCAV:DIAG0:11:MODECFG is in the Disable state. The use case would then be just read only for the TCAV PVs

lcls_tools/common/devices/tcav.py
```diff
@@ -0,0 +1,228 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super(TCAVMetadata, self).__init__(*args, **kwargs)
+
+
+class TCAV(Device):
+    controls_information: SerializeAsAny[TCAVControlInformation]
+    metadata: SerializeAsAny[TCAVMetadata]
+
+    def __init__(self, *args, **kwargs):
```
kabanaty, line 74: I don't think you need an `__init__` function when using Pydantic

kabanaty (commented):
Looks good in general. Some cleanup to be done with implementing Pydantic validation in some of the TCAV properties.

lcls_tools/common/devices/tcav.py
phys-cgarnier: Sorry, the expected type is actually an enum_str, I couldn't get the validation to work how I thought I could.
Unless I made the properties pydantic objects

lcls_tools/common/devices/tcav.py
```diff
@@ -0,0 +1,242 @@
...
+            TimeoutError: If the PV does not return control variables within the timeout period.
+        """
+        mode_config_options = self.PVs.mode_configs.get_ctrlvars(timeout=2.5)
+        if not mode_config_options:
+            raise TimeoutError(
+                "Timeout while retrieving control variables from mode_configs PV."
+            )
+
+        self._mode_config_options.update(
+            {option: i for i, option in enumerate(mode_config_options["enum_strs"])}
```
eloise-nebula, line 69: I like this pattern better than the old one, nice.

lcls_tools/common/devices/tcav.py
```diff
@@ -0,0 +1,242 @@
...
+            @wraps(setter_method)
+            def decorated(self, enum_str):
+                if not isinstance(enum_str,str):
+                    raise TypeError(f"{enum_str} is not of type: str")
+                if enum_str not in field_options:
+                    raise ValueError(f"{enum_str} not in list of acceptable enumerate string PV values")
+                return setter_method(self, enum_str)
+            return decorated
+        return decorator
+    """
```
eloise-nebula, line 146: This can be removed I think.

lcls_tools/common/devices/tcav.py
eloise-nebula: I was wrong about this, the annotations require you to wrap your functions with another decorator. Your current pattern of checking in place is fine.

lcls_tools/common/devices/tcav.py
```diff
@@ -0,0 +1,242 @@
...
+            __context: Reserved for Pydantic internals. Must be present for compliance.
+        Raises:
+            TimeoutError: If any PV fails to return its control variables.
+        """
+        _ = __context  # avoid linter warning for unused variable
+        self.set_mode_config_option()
+        self.set_amplitude_feedback_options()
+        self.setup_phase_feedback_option()
+
+    def set_mode_config_option(self):
```
eloise-nebula, line 54: Maybe this should be called `get_mode_config_option`?

lcls_tools/common/devices/tcav.py
```diff
@@ -0,0 +1,242 @@
...
+            )
+
+        self._amplitude_feedback_options.update(
+            {
+                option: i
+                for i, option in enumerate(amplitude_feedback_options["enum_strs"])
+            }
+        )
+
+    def setup_phase_feedback_option(self):
```
eloise-nebula, line 93: This should be consistent with the above `set_amplitude_feedback_option` too I think.

eloise-nebula (approved):
Hi Chris, thanks for the thoughtful update. I like the new pattern for checking the amplitude and phase feedback options a lot better.

I was wrong when I said the validation should just require annotations like this `def foo(bar: int)`. They need a decorator too, and it looks like we only do that in one place in our code: https://github.com/slaclab/lcls-tools/blob/fbaaa7f0378010ed1359891371bbd6c177880599/lcls_tools/common/devices/magnet.py#L38-L40
If you want to do validation still, I would suggest you just throw `ValueError` in the appropriate places. Either that or remove the silent failures.

Thank you!

PR #271, opened 2025-05-08
Bayesian Refactor

eloise-nebula:
My proposed refactor of `lcls_tools.common.data.fit`. This PR covers `method_base.py` and `methods.py`. If this goes through, I can update with a change to projection fit.

#### `lcls_tools.common.model.optimize`

This module contains the relevant fitting methods. I copied `max_likelihood` from `lcls_tools.common.data.fit`. MLE minimizes the subtraction of two factors, the forward minus the prior. The forward looks like `Σ (y - f(params, x))^2`over all data points. This is the same as least squares. In this case, the prior operates as a penalty function that punishes the fit from deviating too far from an initial guess. For instance, the prior penalty function for the mean looks like a Gaussian. By subtracting the prior from the forward, the algorithm prefers fits that are close to the initial guess for the mean.

`max_likelihood` takes the forward and the prior as input directly. `param_fit` performs a `max_likelihood` fit using the forward and a list of Parameter classes.

#### `lcls_tools.common.model.gaussian`

An example of how to use `lcls_tools.common.optimize`. It sets up its own parameters with `optimize.Parameter` classes, then assigns them to a list accessible within the module. Then it wraps `lcls_tools.common.optimize.param_fit` in its own fit method.

The advantage of this method over the `lcls_tools.common.data.fit` method is that the user call `lcls_tools.common.model.gaussian.fit` directly without having to initialize a model object first. I'm definitely open to suggestions, I think this code is a little rough in spots. Don't consider any of the names to be set in stone. I expect everything to get moved around once we do our big renaming later this month.


lcls_tools/common/model/gaussian.py
```diff
@@ -0,0 +1,85 @@
...
+class offset(optimize.Parameter):
+    name = "off"
+    bounds = (0, 1)
+    @staticmethod
+    def init(x, y):
+        return y.min() + 0.01
+    
+    @staticmethod
+    def prior(off, off_0):
+        return norm(off_0, 0.5).logpdf(off_0)
```
kabanaty, line 74: This method takes `off` and `off_0` as arguments, but only ever uses `off_0`.  Should this line read `return norm(off_0,  0.5).logpdf(off)`?

lcls_tools/common/model/gaussian.py
```diff
@@ -0,0 +1,85 @@
...
+    
+class amplitude(optimize.Parameter):
+    name = "amp"
+    bounds = (0, 1)
+    @staticmethod
+    def init(x, y):
+        return y.max() - y.min() - 0.01
+    
+    @staticmethod
+    def prior(amp, amp_0):
```
kabanaty, line 53: Is this function sensitive to an `amp_0` of 0? Perhaps there should be a catch to force `amp_0 < 0`.

lcls_tools/common/model/optimize.py
```diff
@@ -0,0 +1,107 @@
...
+def param_fit(curve, params, pos, data, use_prior=False):
+    """
+    Given a curve function and parameter objects, computes a 2D MLE fit on
+    normalized data.
+
+    Arguments:
+        curve (Callable[x, params]): The curve to be fit.
+            x: The data positions.
+            params: A list of parameters scipy.optimipze.minimize will fit.
+        params (list[Parameters]): A list of Parameter objects. Indicies
```
kabanaty, line 52: Type hint should be `(list[Parameter])` to match the class name.

lcls_tools/common/model/optimize.py
```diff
@@ -0,0 +1,107 @@
...
+
+def max_likelihood(curve, prior, x, y, init, bounds=None, use_prior = False):
+    """
+    Computes a 2D curve fit using Maximum Likelihood Estimation (MLE).
+
+    Arguments:
+        curve (Callable[x, params]): The curve to be fit.
+            x: The data positions.
+            params: A list of parameters scipy.optimize.minimize will fit.
+        prior (Callable[x, params]): The penalty function that encourages
```
kabanaty, line 79: `prior` doesn't get called with `x`, only `params`

lcls_tools/common/model/optimize.py
```diff
@@ -0,0 +1,107 @@
...
+        params (list[Parameters]): A list of Parameter objects. Indicies
+              correspond to the params arguments in curve and prior.
+        pos (np.array[float]): The data positions.
+        data (np.array[float]): The data weights.
+        use_prior (bool): Flag to apply the prior penalty in MLE fit.
+    """
+    x = (pos - np.min(pos)) / (np.max(pos) - np.min(pos))
+    y = (data - np.min(data)) / (np.max(data) - np.min(data))
+    init = [p.init(x, y) for p in params]
+    print(init)
```
kabanaty, line 61: Looks like this was leftover from debugging.

lcls_tools/common/model/super_gaussian.py
```diff
@@ -0,0 +1,31 @@
...
+
+def curve(x, mean, sigma, amp, off, n):
+    exp = abs((x - mean) / sigma) ** (n / 2)
+    exp = 2 ** (1 / 2 - n / 4) * exp
+    A = np.sqrt(2 * np.pi) * amp
+    return A * norm.pdf(exp) + off
+
+class order(optimize.Parameter):
+    name = "n"
+    bounds = (0, None)
```
kabanaty, line 15: Will no upper bound provide a good fit?

lcls_tools/common/model/optimize.py
```diff
@@ -0,0 +1,107 @@
...
+
+    Out:
+        res: scipy OptimizeResult object.
+    """
+    if not use_prior:
+        def prior(p):
+            return 0
+    def loss(params, x, y):
+        v = np.sum((y - curve(x, params)) ** 2)
+        v -= prior(params)
```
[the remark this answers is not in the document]
eloise-nebula: `a -= 1` is equivalent to `a = a - 1`.

lcls_tools/common/model/super_gaussian.py
eloise-nebula: The order is a unitless parameter that controls how square the profile looks. It's clear the order shouldn't be negative, but there isn't a physical restraint on the order like there is for the mean. The mean gives us the position of our fit on a normalized x axis from 0 to 1. This algorithm won't fit outside of those boundaries, meaning any gaussian profile must be roughly centered in the measurement's window. The upper boundary for the order doesn't have similar physical meaning, and if it does it would probably depend on the application.

It's work when I've tried it. And this is the same boundary condition that the matlab code uses.

lcls_tools/common/model/gaussian.py
eloise-nebula: I'll remove the `-0.01` term from the amplitude init. Then this will only happen when `y.max() == y.min()`. Hopefully that won't come up so often.

lcls_tools/common/model/gaussian.py
```diff
@@ -0,0 +1,90 @@
...
+    name = "mean"
+    bounds = (0, 1)
+
+    @staticmethod
+    def init(x, y):
+        return np.average(x, weights=y)
+
+    @staticmethod
+    def prior(mean, mean_0):
+        return norm(mean_0, 0.1).logpdf(mean)
```
eloise-nebula, line 21: Why does this prior use the 0.1 factor here? Does it have any physical or statistical significance?

lcls_tools/common/model/gaussian.py
```diff
@@ -0,0 +1,90 @@
...
+    name = "sigma"
+    bounds = (1e-10, None)
+
+    @staticmethod
+    def init(x, y):
+        return np.sqrt(np.cov(x, aweights=y))
+
+    @staticmethod
+    def prior(sigma, sigma_0):
+        return gamma(2.5, loc=0, scale=1 / 5.0).logpdf(sigma)
```
eloise-nebula, line 39: I copied this directly from `lcls_tools.common.data.fit.methods`. I think it must be a typo or a copy error. We compute a first guess for the std, but we never use it. Using a static guess would defeat the purpose here.

lcls_tools/common/model/gaussian.py
```diff
@@ -0,0 +1,90 @@
...
+    name = "amp"
+    bounds = (0, 1)
+
+    @staticmethod
+    def init(x, y):
+        return y.max() - y.min()
+
+    @staticmethod
+    def prior(amp, amp_0):
+        var = 0.05
```
eloise-nebula, line 57: Here's another place with an arbitrary factor. Does this have statistical or physical significance?

lcls_tools/common/model/super_gaussian.py
```diff
@@ -0,0 +1,35 @@
...
+
+    @staticmethod
+    def init(x, y):
+        return 2
+
+    @staticmethod
+    def prior(n, n_0):
+        return 0
```
eloise-nebula, line 24: I didn't write a prior here because I'm not sure how to solve for or generate one.

eloise-nebula (commented):
I left some of my own comments because I copied code I'm not familiar with. @roussel-ryan could you send me a link to a textbook or a paper that derives these priors? I'd like to include a reference in the docs for the priors and where they come from.

lcls_tools/common/model/gaussian.py
```diff
@@ -0,0 +1,90 @@
...
+        return mean * x_scale + min(x)
+
+
+class sigma(optimize.Parameter):
+    name = "sigma"
+    bounds = (1e-10, None)
+
+    @staticmethod
+    def init(x, y):
+        return np.sqrt(np.cov(x, aweights=y))
```
nneveu, line 35: What's the difference between `aweights` here and `weights` above?

lcls_tools/common/model/gaussian.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @staticmethod
+    def prior(sigma, sigma_0):
+        return gamma(2.5, loc=0, scale=1 / 5.0).logpdf(sigma)
+
+    @staticmethod
+    def denormalize(sigma, x, y):
+        x_scale = np.max(x) - np.min(x)
```
nneveu, line 43: Is normalization different for sigma? This doesn't match the equation above? Also, same question about the `y` here.

lcls_tools/common/model/gaussian.py
```diff
@@ -0,0 +1,90 @@
...
+    @staticmethod
+    def prior(amp, amp_0):
+        var = 0.05
+        alpha = (amp_0**2) / var
+        beta = amp_0 / var
+        return gamma(alpha, loc=0, scale=1 / beta).logpdf(amp)
+
+    @staticmethod
+    def denormalize(amp, x, y):
+        y_scale = np.max(y) - np.min(y)
```
nneveu, line 64: `x` is provided but not used here. Same q's as above.

lcls_tools/common/model/gaussian.py
```diff
@@ -0,0 +1,90 @@
...
+        return amp * y_scale
+
+
+class offset(optimize.Parameter):
+    name = "off"
+    bounds = (0, 1)
```
nneveu, line 70: `bounds` are defined but not used?

lcls_tools/common/model/gaussian.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @staticmethod
+    def init(x, y):
+        return y.min()
```
nneveu, line 74: Assumes we want y offset? I think x and y are being used in difference contexts, and it's a bit unclear.

lcls_tools/common/model/gaussian.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @staticmethod
+    def prior(off, off_0):
+        return norm(off_0, 0.5).logpdf(off)
+
+    @staticmethod
+    def denormalize(off, x, y):
+        y_scale = np.max(y) - np.min(y)
```
nneveu, line 82: Same discrepancies as above.

lcls_tools/common/model/optimize.py
```diff
@@ -0,0 +1,111 @@
...
+    @abstractmethod
+    def prior(par, par_0): ...
+
+    @staticmethod
+    @abstractmethod
+    def denormalize(par, x, y): ...
+
+
+def param_fit(curve, params, pos, data, use_prior=False):
+    """
```
nneveu, line 43: Lots of inputs w/o forcing type or checks. I'd be concerned about things getting mixed up in these calls.

lcls_tools/common/model/optimize.py
```diff
@@ -0,0 +1,111 @@
...
+              correspond to the params arguments in curve and prior.
+        bounds (tuple(tuple[float, float])): Boundaries for the fitted params.
+        use_prior (bool): Flag to apply the prior penalty.
+
+    Out:
+        res: scipy OptimizeResult object.
+    """
+    if not use_prior:
+
+        def prior(p):
```
nneveu, line 96: Why have an input `p` if you're returning 0 always?

lcls_tools/common/model/super_gaussian.py
nneveu: Do we need it then? Or we do bc the ML/optimization code calls for it?

nneveu: Same with denormalize?

lcls_tools/common/model/super_gaussian.py
```diff
@@ -0,0 +1,35 @@
...
+    @staticmethod
+    def denormalize(n, x, y):
+        return n
+
+
+params = gaussian.params + [order]
+
+
+def fit(pos, data):
+    return optimize.param_fit(curve, params, pos, data, use_prior=False)
```
nneveu, line 35: If we're going to call this every time, I will need to look at that the `param_fit` function more closely. Putting a pin in it for myself.

tests/unit_tests/lcls_tools/common/model/gaussian.py
```diff
@@ -0,0 +1,19 @@
+import numpy as np
+import unittest
+from lcls_tools.common.model import gaussian
+
+
+class TestGaussianFits(unittest.TestCase):
```
nneveu, line 6: We should add a test that pulls a defined set of data from a file that doesn't change. It's not needed for this PR, just a note.

nneveu (changes requested):
I left a round of comments. I can review again tomorrow. I think the static methods in each class are going back to the copy/paste style that Rob recommended we avoid. I think this is showing up in the normalization differences. Unless we really want to normalize differently in certain places?  Thanks for your hard work!

nneveu:
Also, very nice PR note! We should all be like you :)

eloise-nebula:
Hi @nneveu, you left a few comments on apparently extraneous arguments and normalization, so I'll answer them broadly in two comments.

The purpose of the `Parameter` classes is to provide an easy way to store functions and configurations related to a fitted parameter. For instance, we may want to calculate a first guess for each parameter. We accomplish that here: https://github.com/slaclab/lcls-tools/blob/56cdef90033fe59088be2fb4f244602eb89cbd6b/lcls_tools/common/model/optimize.py#L59 Where `x` and `y` are the data to be fit, and `params` is a list of `Parameter` objects. This line expects every `Parameter.init` function to take `x` and `y` as arguments, regardless of whether or not those arguments are used. That's why we have apparently unused arguments throughout, such as in this `amplitude.init` function: https://github.com/slaclab/lcls-tools/blob/56cdef90033fe59088be2fb4f244602eb89cbd6b/lcls_tools/common/model/gaussian.py#L51-L53 We later solve for our prior penalty function using `Paremter.prior` in the same manner: https://github.com/slaclab/lcls-tools/blob/56cdef90033fe59088be2fb4f244602eb89cbd6b/lcls_tools/common/model/optimize.py#L64-L65 This method is compact, easy to read, and maximizes code sharing. Our `Parameter.prior` and `Parameter.init` functions live in the same place in our code, whereas this information is scattered throughout `methods.py`. The `Parameter` abstract class at the top of `optimize.py` defines the expected arguments for `init` and `prior`.

eloise-nebula:
In `projection.py`, we normalize our `x` and `y` data such that they extend between 0 and 1, here: https://github.com/slaclab/lcls-tools/blob/3e0121f333f7414a9a48379723de2ee2472b5394/lcls_tools/common/data/fit/projection.py#L35-L42 and here: https://github.com/slaclab/lcls-tools/blob/3e0121f333f7414a9a48379723de2ee2472b5394/lcls_tools/common/data/fit/projection.py#L77 There are two reasons for this as far as I can tell. For one, it lets us set solid boundary conditions for our fit. If we know the data lies between 0 and 1, we know that our gaussian fit can't have an amplitude greater than one. 

The other reason is it lets us define our prior functions independently from the scale of the data. Right now, we define the prior for the mean to be `norm(mean_0, 0.1).logpdf(mean)`. If `mean, mean_0 >> 0.1`, this prior will strongly penalize any result that isn't right on top of `mean_0`. We can avoid this issue by normalizing the data.

Each parameter gets denormalized differently. Let's look at the exponent for our gaussian, `(x - mu)^2 / sigma^2`. When we normalize our `x` data, we displace it by a factor equal to `min(x)`. For `mu` to match, `mu` must be displaced by the same factor. Then, to scale `x` such that it lies between 0 and 1, we divide it by `max(x) - min(x)`. For `mu` and `sigma` to match, they must be scaled by the same factor. `sigma` is just a scale factor, and does not have to be displaced by `min(x)`. Therefore they require different denormalization functions.

lcls_tools/common/model/gaussian.py
eloise-nebula: From the `np.average` docs:
> weights, array_like, optional: An array of weights associated with the values in a. Each value in a contributes to the average according to its associated weight.

From the `np.cov` docs:
> aweight, sarray_like, optional: 1-D array of observation vector weights. These relative weights are typically large for observations considered “important” and smaller for observations considered less “important”.

In short, they're the same.

roussel-ryan:
Additionally, normizing the projection is important for fitting robustness across different signals (projections of images, wires, etc) to maintain gradients near unity.

lcls_tools/common/model/gaussian.py
eloise-nebula: Bounds are used here: https://github.com/slaclab/lcls-tools/blob/56cdef90033fe59088be2fb4f244602eb89cbd6b/lcls_tools/common/model/optimize.py#L67

lcls_tools/common/model/gaussian.py
eloise-nebula: That's fair. I've been using `pos` and `data` for the un-normalized data. I'll make that consistent throughout.

lcls_tools/common/model/optimize.py
eloise-nebula: I'm happy to include annotations, but I recently realized they don't do anything on their own. Python doesn't do any type checking at runtime. I think we got confused because Pydantic does do type checking at runtime. We should add a type checker to our github workflow, but in general annotations don't stop people from making mistakes like:
```
foo(a: int):
    return a + 1
foo("whoops!")
```

robnagler: Annotations are expensive for these types of interfaces (where the arguments are generally the same e.g. float), and in the given example, calling `foo("whoops!")` will generate an error at run-time. 

If you want to put your development dollars towards something with long term value, write tests that validate boundary conditions and error cases that aren't caught at run-time. Tests have the advantage that they serve as executable documentation of how the interface is supposed work (or not). Annotations do not do this. They simply tell you what you can pass, which as noted above, are often obvious.

lcls_tools/common/model/gaussian.py
nneveu: Sorry, should have looked this up myself!

nneveu:
Took a second quick pass at this. I should have been less sloppy with my comments above. I'm familiar with normalization ([especially min-max scaling to [0 to 1] ](https://en.wikipedia.org/wiki/Normalization_(statistics))) for optimization, fitting, etc. What I was missing here is what I think of as scaling vs. boundary normalization, both of which could be called normalization. @eloise-nebula I'll ping you for a quick chat. I'm still not a fan of all the repeated code/unused inputs, but I definitely could spend more time looking at it/understanding. We (I mainly mean _I_) need to do a small write up on the math here, what we want, with sources, etc. I'll add that to my to-do. In the mean time, I'll let @roussel-ryan get his review in.

nneveu (approved):
Thanks for going over the code with me in detail!

nneveu:
@roussel-ryan please review this before we merge/close.

roussel-ryan:
Yes, sorry for the delay, I will get to this in the next 12-16 hrs

lcls_tools/common/model/optimize.py
```diff
@@ -0,0 +1,111 @@
...
+        v = np.sum((y - curve(x, params)) ** 2)
+        v -= prior(params)
+        return v
+
+    res = scipy.optimize.minimize(
+        loss,
+        init,
+        args=(x, y),
+        bounds=bounds,
+        method="Powell",
```
roussel-ryan, line 109: you might want to let the method be selected by scipy unless there is a reason to choose Powell explicitly

lcls_tools/common/model/gaussian.py
roussel-ryan: We chose a value of 0.1 (10% of the signal domain) based on empirical testing

nneveu: Do you have more info on how the empirical value was picked? Maybe paper link or tech note / data summary, etc.

lcls_tools/common/model/gaussian.py
nneveu: Same here^

lcls_tools/common/model/optimize.py
```diff
@@ -0,0 +1,111 @@
...
+        return np.sum([p.prior(v, i) for p, v, i in zip(params, vec, init)])
+
+    bounds = tuple(p.bounds for p in params)
+    res = max_likelihood(forward, prior, x, y, init, bounds, use_prior)
+    return {p.name: p.denormalize(val, pos, data) for p, val in zip(params, res.x)}
+
+
+def max_likelihood(curve, prior, x, y, init, bounds=None, use_prior=False):
+    """
+    Computes a 2D curve fit using Maximum Likelihood Estimation (MLE).
```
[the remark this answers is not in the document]
nneveu: Can you add this how you would like to see it?

nneveu:
In response to the general comments:
> 
>     * can we put the new files inside something like `common/model/projection/` or similar? Model by itself seems not descriptive enough

This could be used for something aside from image projection...

>     * I'm not seeing how these changes are being incorporated into the greater package, ie. measurement classes etc. is this a future PR or can you point me to the PR that made those changes?
That scope is not included here and would need to be another PR.

lcls_tools/common/model/gaussian.py
roussel-ryan: We don't have any formal data testing / write up, but could add in the near future. This was refined after multiple MD shifts in a causal fashion

roussel-ryan: I wouldn't say that its tuned to emittance measurements, its tuned to find gaussian profiles in noisy data

nneveu: I think the point is not all data is profile data. Let's go with Eloise's solution, and you can submit a follow up PR if that's not working for the ML applications.

roussel-ryan: right, but if you are trying to find a gaussian in a signal, regardless of where that signal came from then this should work the best

roussel-ryan: we can subclass if there are specific instances where this doesn't work

nneveu: This may be true, but we don't have the reference/data to back that up? I don't want to merge hard coded empirical constants without something to point to as a good reason for doing so. We could revisit after the references are there.

roussel-ryan: that's fair, can we wait to merge until we have that data? or for now have the prior functions be uniform / remove the prior methods?

eloise-nebula: I'd like to merge soon, and I'll leave the prior functions uniform for now.

roussel-ryan: ok, works for me, just make sure that they are wide enough to cover the expected domain of parameter values

roussel-ryan: changes look good to me

PR #272, opened 2025-05-09
Wire Scan Bug Fixes: Device List Cleanup, Optional Fields for Validation, and Redundant Logic Removal

kabanaty:
This PR addresses several issues identified during machine testing and review:

**Device metadata cleanup:**

- Removed non-existent devices from the LTUS LBLM list
- Corrected device name from LBLM32A to LBLMS32A

**Pydantic model fixes:**

- Updated `TMITLoss` and `WireBeamProfileMeasurement` to set `idx_before`, `idx_after`, `bpms`, and `my_buffer` as `Optional`, resolving validation errors during instantiation

**Refactored internal method calls:**

- Removed redundant calls to `reserve_buffer()` and `create_device_dictionary()` after moving those operations into `@model_validator` blocks
- Replaced argument-based calls to `scan_with_wire()` and `get_bsa_data()` with class attribute versions

**Function argument corrections and typo fixes:**

- Fixed typos like `startsWith` → `startswith`
- Updated method signatures and logic to align with attribute-based buffer/device access

These changes improve robustness during initialization, reduce duplicated work, and ensure device lists accurately reflect real hardware.


lcls_tools/common/measurements/wire_scan.py
```diff
@@ -57,6 +57,15 @@ class WireBeamProfileMeasurement(Measurement):
     beam_fit: BaseModel = ProjectionFit
     fit_profile: bool = True
 
+    # Extra fields to be set after validation
+    my_buffer: Optional[edef.BSABuffer] = None
+    devices: Optional[dict] = None
+
+    @model_validator(mode="after")
+    def run_setup(self) -> "WireBeamProfileMeasurement":
```
eloise-nebula, line 65: This function annotation says you're outputting a "WireBeamProfileMeasurement", but there's no return statement here. Also, [Pydantic docs](https://docs.pydantic.dev/latest/concepts/validators/#model-after-validator) say you can annotate like this:
```
from typing_extensions import Self

...

@model_validator(mode="after")
def run_setup(self) -> Self:
```

lcls_tools/common/measurements/wire_scan.py
```diff
@@ -57,6 +57,15 @@ class WireBeamProfileMeasurement(Measurement):
...
+        self.my_buffer = self.reserve_buffer()
```
eloise-nebula, line 66: The Pydantic constructor will allow users to initialize `my_buffer` as a keyword argument. As it's written, function will always overwrite `my_buffer`. I think we should allow users to use their own buffer if they want to.

eloise-nebula: I recognize that this will allow users to mismatch beampaths and edef buffers. I think that's okay?

lcls_tools/common/devices/yaml/metadata.py
```diff
@@ -148,16 +148,7 @@ def get_wire_metadata(wire_names: List[str] = []):
...
-        "PMT246",
-        "PMT430",
-        "PMT431",
-        "PMT550",
-        "PMT755",
-        "PMT756",
-        "PMT820",
-        "PMT850",
-        "LBLM32A",
+        "LBLMS32A",
```
eloise-nebula, line 151: You added this metadata but there's no changes to the YAML files. Maybe rerun YAML generation?

eloise-nebula (commented):
Hi Tyler, just a couple comments. Looks good in general. Let me know if you have thoughts or questions.

lcls_tools/common/measurements/wire_scan.py
kabanaty: Ahh yes, outputting to `Self` is what I was really after.  Let me try that.

lcls_tools/common/measurements/wire_scan.py
kabanaty: I can do an if loop along the lines of `if my_buffer is None: self.my_buffer = self.reserve_buffer()`.  If someone is passing they're own buffer but gets the linac wrong, that seems like an edge case.  I'm happy to let users figure that one out their own.

lcls_tools/common/measurements/wire_scan.py
```diff
@@ -57,6 +58,16 @@ class WireBeamProfileMeasurement(Measurement):
...
+    # Extra fields to be set after validation
+    my_buffer: Optional[edef.BSABuffer] = None
+    devices: Optional[dict] = None
+
+    @model_validator(mode="after")
+    def run_setup(self) -> Self:
+        if self.my_buffer is None:
+            self.my_buffer = self.reserve_buffer()
+        self.devices = self.create_device_dictionary()
+
```
eloise-nebula, line 70: Pydantic says:
> Important note: the validated instance should be returned. 
You need to `return self` at the end here.

eloise-nebula (approved):
Thanks for the changes. All ready to merge once tests are passing.

PR #275, opened 2025-05-14
Extend Area model to support BPMs and LBLMs

kabanaty:
This PR extends the `Area` model to support `BPM` and `LBLM` devices which were previously unsupported.  Attempting to create and `Area` containing these devices would result in Pydantic validation errors due to missing model fields.

The changes introduce:

- A `bpm_collection` field (aliased from `bpms`) using the `BPMCollection` class
- A `lblm_collection` field (aliased from `lblms`) using the `LBLMCollection` class
- Field validators to properly initialize these collections from YAML dictionaries

With this update, BPM and LBLM devices are now handled consistently alongside existing device types such as magnets, wires, and screens.

### DIFF
--- a/lcls_tools/common/devices/area.py
+++ b/lcls_tools/common/devices/area.py
@@ -249,7 +249,7 @@ class Area(lcls_tools.common.BaseModel):
         self,
         wire_name: str = None,
     ) -> bool:
-        return screen_name in self.wires
+        return wire_name in self.wires
 
     def does_bpm_exist(
         self,
--- a/lcls_tools/common/devices/area.py
+++ b/lcls_tools/common/devices/area.py
@@ -107,46 +107,45 @@ class Area(lcls_tools.common.BaseModel):
         mode="before",
     )
     def validate_magnets(cls, v: Dict[str, Any]):
-        if v:
-            # Unpack the magnet data from yaml into MagnetCollection
-            # before creating the magnet_collection
-            return MagnetCollection(**{"magnets": {**v}})
+        if not v:
+            return None
+        return MagnetCollection(magnets=v)
 
     @field_validator(
         "screen_collection",
         mode="before",
     )
     def validate_screens(cls, v: Dict[str, Any]):
-        if v:
-            # Unpack the screens data from yaml into ScreenCollection
-            return ScreenCollection(**{"screens": {**v}})
+        if not v:
+            return None
+        return ScreenCollection(screens=v)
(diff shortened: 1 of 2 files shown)


lcls_tools/common/devices/area.py
```diff
@@ -104,6 +130,24 @@ def validate_wires(cls, v: Dict[str, Any]):
...
             return WireCollection(**{"wires": {**v}})
 
+    @field_validator(
+        "bpm_collection",
+        mode="before",
+    )
+    def validate_bpms(cls, v: Dict[str, Any]):
+        if v:
+            # Unpack the bpms data from yaml into BPMCollection
+            return BPMCollection(**{"bpms": {**v}})
```
eloise-nebula, line 140: If v is supposed to be a dictionary then `{**v} == v` in general. This line could read `BPMCollection(bpms=v)`. But maybe there's a particular reason for this I'm not seeing? Matt wrote the original one.

kabanaty:
> Hi Tyler, thanks for the PR. You mentioned to me over slack that these methods could replace your recent `load_full_beampath` PR. Could you remove that function here?

Yes I'll commit some changes to remove that function.

lcls_tools/common/devices/area.py
kabanaty: I can change all the `validation_<devices>` functions to something like:

`if not v:`
`    return None` 
`return BPMCollection(bpms=v)`

eloise-nebula: Oh, I see. I think that would be more clear.

kabanaty:
After some MD time, this isn't the right solution.  I'll make a separate PR for the `load_full_beampath` change.

lcls_tools/common/devices/area.py
```diff
@@ -81,28 +107,45 @@ def __init__(
         mode="before",
     )
     def validate_magnets(cls, v: Dict[str, Any]):
-        if v:
```
nneveu, line 84: Was the sytnax here and w/ the screen below causing errors? This now assumes there is nothing to unpack. Is that always true, or we want to force that to be true?

nneveu (approved):
I have two open questions here, but will go ahead and approve. Thanks for adding additional comments and adding bpms/lblms!

nneveu:
Sorry just now seeing the discussion above. I will go ahead and resolve my comments and merge.

PR #279, opened 2025-05-19
Filter Bad Rows

eloise-nebula:
`lcls_elements.csv` has a lot of bad rows. It's annoying to have to clean these up every time we run something. This PR allows us to filter out rows if their columns contain a bad prefix. My main goal here was to filter out the "*" and "\t-No Area-" rows, but I made it general in case we find another bad prefix. Also works for exact matches. All the data is loaded as strings to start so I think this method is okay. We could possibly switch to regex, but I'm worried that it would get too general too quickly. Easy change though.


lcls_tools/common/devices/yaml/generate.py
```diff
@@ -43,20 +46,35 @@ def __init__(
...
+            elements = list(filter(_is_filtered_row, csv_reader))
 
             # make the elements from csv stripped out with only information we need
             def _is_required_field(pair: tuple):
-                key, _ = pair
+                key, value = pair
                 return key in required_fields
 
             # only store the required fields from lcls_elements, there are lots more!
             elements = [
```
nneveu, line 73: I'd rather rename 'elements' either on line 65 or here. I think technically this could work fine, but makes me nervous.

nneveu (approved):
Great small PR! Only change I would like/suggest is the elements name. After that we can merge. Thanks!

PR #280, opened 2025-05-27
Modularize emittance calc for generalization.

dylanmkennedy:
Modularizes emittance calculation for generalization to multi-device measurements (e.g. 3-wire measurement).

### DIFF
--- a/lcls_tools/common/data/emittance.py
+++ b/lcls_tools/common/data/emittance.py
@@ -9,7 +9,7 @@ from lcls_tools.common.data.model_general_calcs import (
 )
 
 
-def compute_emit_bmag(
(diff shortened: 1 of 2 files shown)


roussel-ryan:
So I would agree with that Shamin, but I'm concerned about connecting it too closely to the emittance measurement objects so that these calcs can be used outside the context of slac-tools. In that case maybe it makes sense to include these convenience functions in this file?

shamin-slac:
@dylanmkennedy were you able to confirm or test that compute_emit_bmag works for the multi-device case?

shamin-slac (approved):
Looks good to me!

PR #281, opened 2025-05-27
Generate Quality of Life

eloise-nebula:
A couple of things about running write.py were bothering me. Here's the fixes for them. I fixed one bug where a variable `pv_info` wouldn't be initialized on failure to connect. `write.py` now allows the user to choose what devices they want to write to.


eloise-nebula:
This code doesn't work the way I expected it to. I'm going to come up with something totally new.

PR #282, opened 2025-06-02
write.py Command Line Interface

eloise-nebula:
I wrote this PR because I was annoyed at how `write.py` reads over every device in every single area every single time. The task turned out to be a little bit more complicated than I thought, but I think this is a good solution.

This PR mainly adds two new features, along with a CLI and a few QOL changes.

`write.py` now has three modes. Each mode pulls data from `lcls_elements.csv`, but each one applies that data differently.
- `overwrite` is the default, and works the same as before. The data is treated as a complete source of information, and each YAML file it writes is completely overwritten.
- `greedy` treats the data as a complete update. It adds new fields and devices while overwriting old ones where old fields exist. 
- `lazy` treats the data as a partial update. It only adds new fields and devices, and leaves anything already there alone.

`write.py` also now supports selecting the devices to read and write from. If you use this mode with `overwrite` you will erase the old devices from the YAML files. Only use this feature with the `greedy` or `lazy` write modes.

`python write.py` now supports command line arguments, `--mode` and `--devices`, which should be self explanatory. This PR also adds tests for `write.py`.

### DIFF
--- a/lcls_tools/common/devices/yaml/write.py
+++ b/lcls_tools/common/devices/yaml/write.py
@@ -3,6 +3,7 @@ import os
 from lcls_tools.common.devices.yaml.generate import YAMLGenerator
 from typing import Optional, List, Dict
 import collections.abc
+import argparse
 
 
 class YAMLWriter:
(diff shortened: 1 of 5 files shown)


lcls_tools/common/devices/yaml/write.py
```diff
@@ -15,70 +20,125 @@ def areas(self) -> List[str]:
...
+            else:
+                if k not in target:
+                    target[k] = v
+        return target
+
+    def greedy_write(
+        self, area: Optional[str] = "GUNB", devices: List[str] = None
+    ) -> None:
+        current = self._get_current(area)
+        update = self._constuct_yaml_contents(area=area, devices=devices)
```
kabanaty, line 92: If `_construct_yaml_contents` returns `None`, then `_greedy_update` will break. Same for this equivalent line in `lazy_write` and `_lazy_update`.

lcls_tools/common/devices/yaml/write.py
```diff
@@ -15,70 +20,125 @@ def areas(self) -> List[str]:
     def _is_area(self, area: str) -> bool:
         return area in self.generator.areas
 
-    def _constuct_yaml_contents(self, area: str) -> Dict[str, str]:
+    def _constuct_yaml_contents(
```
kabanaty, line 23: Looks like this is a legacy typo, but "construct" is misspelled. Missing the r!

eloise-nebula:
@kabanaty thanks for the comments. I think I've addressed them all.

nneveu (approved):
Thanks for the QoL update!

PR #284, opened 2025-06-04
Update optimize.py

roussel-ryan:
Refactored `maximum_likelihood` function into `map_fit` function which optionally takes a prior callable object instead of requiring it (removes use_prior flag). Updates `curve_fit` method to use new function + call structure. Updates docstrings for both functions and fixes minor errors with defining / calling prior functions. Removes default use of "Powell" in `scipy.optimize.minimize`


eloise-nebula (approved):
Thanks for the clarifications. I'll go ahead and merge this.

PR #285, opened 2025-06-04
Fix tcav bugs

phys-cgarnier:
Modified the code in tcav so that epics.PV.get now has the argument use_monitor=False, some users need this flag to pull the the real value of the PV in jupyter notebooks since pyepics has a cached value it can return.

Also for large images and slow servers screen.image can timeout. I increased this timeout to avoid a spew of timeout warnings. (we can discuss this and the duration of the timeout)


nneveu:
Meeting notes - sometimes getting cached values when getting PVs?

nneveu (changes requested):
Remove timeout from screen update. Then this can merge.

PR #291, opened 2025-06-12
Screen Orientation

eloise-nebula:
Adds orientation for relevant devices to our screen object. In the profmon_gui, we pull the orientation information down from the PVs every time. Instead of doing that, I pull the orientation information into our yaml files. This won't be convenient for doing checkouts, but I think we can come up with another solution to that problem. e.g. a checkout mode. Once the checkout is done we shouldn't need to change the orientation again. I'm also uploading profmon_elements.csv, which contains all the information in profmon_gui. We don't create cameras with this data right now, but we probably will in the future.

### DIFF
--- a/lcls_tools/common/devices/screen.py
+++ b/lcls_tools/common/devices/screen.py
@@ -102,10 +102,10 @@ class Screen(Device):
         i = self.controls_information
         pv_cache = getattr(i, "pv_cache", None)
         if pv_cache is not None and not self.new_orientation:
-            if (v := getattr(pv_cache, "orient_x", None)) is not None:
-                return v
-        if (pv := getattr(i, "orient_x", None)) is not None:
-            return pv.get()
+            if "orient_x" in pv_cache:
+                return pv_cache["orient_x"]
+        if (pv := getattr(i.PVs, "orient_x", None)) is not None:
+            return pv.get(as_string=True)
         return None
 
     @property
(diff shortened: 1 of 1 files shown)


lcls_tools/common/devices/yaml/controls_information.py
```diff
@@ -24,11 +25,15 @@ def get_screen_controls_information(screen_names: List[str] = None):
...
     # Stuff like Device-Position mappings for motor/ladder-based screens
-    if screen_names:
-        raise NotImplementedError(
-            "No method of getting additional controls_information for screens."
-        )
-    return {}
+    controls_information = {}
+    for k, v in screen_information.items():
+        pvs = v["controls_information"]["PVs"]
+        if "orient_x" in pvs and "orient_y" in pvs:
```
robnagler, line 31: I think these cached values should be in their own section, e.g. `controls_information.PV_cache`, so the "get" code could look like (`accessor` is the logical name, that is, "key" of PVs):

```py
def pv_get(self, accessor):
    i = self.controls_information 
    if (v := getattr(i["PV_cache"], accessor) is not None:
        return v
    return caget(i["PVs"][accessor])
```

Without the extra scope, there's a name space conflict with accessor names and `PVs` and `control_name` as well as complexity in the definitions in pydantic (wouldn't allow `PVValueSet` to be a type for controls_information).

lcls_tools/common/devices/yaml/generate.py
```diff
@@ -153,6 +152,8 @@ def _construct_information_from_element(
     def _construct_pv_list_from_control_system_name(
         self, name, search_with_handles: Optional[Dict[str, str]]
     ) -> Dict[str, str]:
+        import meme.names
```
kabanaty, line 155: What is this import used for?

lcls_tools/common/devices/yaml/controls_information.py
eloise-nebula: I implemented the PV cache. The main reason the orientation is a PV instead of, say, and IOC feature is so that people can change the orientation in real time while performing camera installs. I added a flag to ignore the cache to support that use case.

robnagler: The cache is fine. I just think the name space should be adjusted.

kabanaty (approved):
Looks good! I'm approving.

PR #301, opened 2025-06-30
Overwrite Mode Works

eloise-nebula:
I recently added a greedy and lazy writer to our YAML generator, but broke the default mode, now called overwrite, in the process. This is a quick three line fix to solve that bug. Thanks to @kabanaty for pointing this out to me.


kabanaty (approved):
Looks great. Thanks Eloise!

Issue #298, opened 2025-06-25
Create scan_utils subdirectory under measurements for shared utilities


kabanaty:
Introduce a new `scan_utils` module under `lcls_tools/common/measurements/` to house shared functionality across diagnostic tools, including wire scans.  This will help reduce duplication and centralize logic such as logging and buffer reservation.

Initial Scope:
- Logging setup
- Buffer reservation utilities
- Placeholder structure for future expansion (e.g. error handling, scan coordination)

Issue #295, opened 2025-06-24
Add support for beam rate PV for Wire devices


kabanaty:
Wire scan measurements require knowledge of the current beam rate, which is available via the `:BEAMRATE` PV. This PV provides information for data acquisition and analysis, particularly when configuring buffers.

Currently, lcls-tools does not interface with this PV. This issue tracks the implementation of beam rate support in relevant wire scan components. Once added, the measurement logic should retrieve and incorporate the beam rate value where applicable.

- Access to :BEAMRATE PV should be implemented in wire device code
- Fallback or error handling is in place if PV is unavailable

kabanaty:
Closed with #299

PR #302, opened 2025-06-30
scan_utils module for shared measurement utilities

kabanaty:
This PR addresses #298 by introducing a new `scan_utils` module under `lcls_tools/common/measurements/`.  The module will house shared functionality used across diagnostic tools like wire scans, reducing code duplication and improving maintainability.

**Changes**

- Created `scan_utils/` directory
- Added initial module `buffer_reservation`
- Updated `wire_scan.py` to use new buffer reservation module


lcls_tools/common/measurements/scan_utils/buffer_reservation.py
nneveu, line 1: If this is the only file in scan_utils, I would rather keep things simpler and just have the buffer_res file in measurements. Especially since we want to move to the new repos soon.

nneveu: I'm also not a fan of 'utils' in file or directory names. If we can name something more informative, I'd rather lean towards that. But I get it's a very common and short name, so it is tempting!

lcls_tools/common/measurements/wire_scan.py
```diff
@@ -3,6 +3,8 @@
 from lcls_tools.common.devices.reader import create_lblm
 from lcls_tools.common.data.fit.projection import ProjectionFit
 from lcls_tools.common.measurements.measurement import Measurement
+from lcls_tools.common.measurements.tmit_loss import TMITLoss
+from lcls_tools.common.measurements.scan_utils.buffer_reservation import reserve_buffer
```
nneveu, line 7: Removing the scan_utils folder would make this import easier, etc. I think it's fine to have a buffer file at the top of measurement.

lcls_tools/common/measurements/wire_scan.py
```diff
@@ -46,9 +46,13 @@ class WireBeamProfileMeasurement(Measurement):
     @model_validator(mode="after")
     def run_setup(self) -> Self:
         if self.my_buffer is None:
-            print("Reserving BSA Buffer...")
-            self.my_buffer = self.reserve_buffer()
-            print(f"Reserved BSA Buffer {self.my_buffer.number}")
+            self.my_buffer = reserve_buffer(
+                beampath=self.beampath,
+                name="LCLS Tools Wire Scan",
+                n_measurements=1600,
```
nneveu, line 52: Where does 1600 come from?

lcls_tools/common/measurements/wire_scan.py
```diff
@@ -166,6 +133,15 @@ def scan_with_wire(self):
         and allows time for the buffer to update its state.
         """
         # Start wire scan
+        if self.my_buffer is None:
```
nneveu, line 136: Is this code repeated from line 49+? Maybe it can also be pulled out if it's general enough to re-use?

nneveu (changes requested):
I think you don't need the extra directory. Otherwise, can prob be merged soon. I had a few other small comments/questions. Thanks!

lcls_tools/common/measurements/wire_scan.py
kabanaty: The value of 1600 buffer points was determined empirically during MD shifts.  The MATLAB GUI currently uses an equally arbitrary n_measurements=2295.  I introduced the beam rate PV (#295) to make more robust speed calculations and buffer management.

nneveu: Are those results written up somewhere for reference?

lcls_tools/common/measurements/scan_utils/buffer_reservation.py
kabanaty: I know that `NDArrayAnnotatedType` class is used all over the place. I don't know about the other two functions.  I would like to move that file. If that file is moved all the import references will need to be updated as well, and that's a potentially breaking change.  If we decide to move forward with a separate directory for `scan_utils`, I'll address `utils.py` in a separate PR.

lcls_tools/common/measurements/wire_scan.py
nneveu: Ok, can we write this up in a technote?

lcls_tools/common/measurements/scan_utils/buffer_reservation.py
kabanaty: Right now, the logging code is just one file. So if we keep everything in `measurements/`, we’ll end up with two additional files. That’s totally fine if we want to keep it simple, I just wanted to flag it in case we’d prefer a cleaner separation later.

eloise-nebula (approved):
Hi Tyler, thanks for writing this great one for the timing system. Just one minor comment. I'll approve, but keep working with Nicole.

lcls_tools/common/measurements/buffer_reservation.py
```diff
@@ -0,0 +1,44 @@
...
+    if logger:
+        logging.info("Reserving buffer...")
+
+    if beampath.startswith("SC"):
+        if destination_mode not in DEST_MODE_MAP:
+            raise BufferError(f"Invalid destination mode: {destination_mode}")
+
+        buf = edef.BSABuffer(name=name, user=user)
+        buf.n_measurements = n_measurements
+        buf.destination_mode = DEST_MODE_MAP[destination_mode]
```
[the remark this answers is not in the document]
kabanaty: I just checked and it appears that it is.  I didn't realize that I could write directly like that.  I just made a new commit addressing this.  Thanks!

lcls_tools/common/measurements/buffer_reservation.py
```diff
@@ -0,0 +1,41 @@
...
+def reserve_buffer(
+    name: str,
+    beampath: str,
+    n_measurements: int,
+    destination_mode: str = "Disable",
+    logger: logging.Logger = None,
+):
+    user = os.getlogin()
+    if logger:
+        logging.info("Reserving buffer...")
```
eloise-nebula, line 19: Do you want to update this PR with your new logging tools?

kabanaty: Oh yes I do.  Nicole found that someone was using the existing logger module, so I made a quick PR to revert and create a new `file_logger`.  I'll update this shortly.

nneveu (approved):

nneveu:
Thanks Tyler!

PR #305, opened 2025-07-08
Update existing logger module

kabanaty:
This PR replaces the existing logger module by including a file handler alongside the existing stream handler.  A custom `sys.excepthook` is also implemented to capture and log all uncaught exceptions. This enables persistent logging for later review and debugging.

### DIFF
--- a/lcls_tools/common/devices/tcav.py
+++ b/lcls_tools/common/devices/tcav.py
@@ -80,7 +80,7 @@ class TCAVControlInformation(ControlInformation):
         Raises:
             TimeoutError: If control variables are not returned within the timeout duration.
         """
-        amplitude_feedback_options = self.PVs.amp_fbenb.get_ctrlvars(timeout=2.5)
+        amplitude_feedback_options = self.PVs.amplitude_fbenb.get_ctrlvars(timeout=2.5)
         if not amplitude_feedback_options:
             raise TimeoutError(
                 "Timeout while retrieving control variables from amp_fbenb PV."


kabanaty:
@eloise-nebula You are correct, the logger is not currently used anywhere.  I wrote the new logger as a proposed scan_util, not realizing we already had a logger directory.  I would like to replace the unused logger with the one that I have written.  I plan on using the new logger in `wire_scan` and also the upcoming wire scan PyDM GUI.

lcls_tools/common/logger/logger.py
```diff
@@ -1,23 +1,38 @@
...
-)
+    def make_handler(handler, use_date_format=False):
+        formatter = logging.Formatter(
+            FORMAT_STRING, datefmt=DATE_FORMAT if use_date_format else None
+        )
+        handler.setFormatter(formatter)
+        return handler
 
+    # File handler
+    logger.addHandler(make_handler(logging.FileHandler(log_file)))
```
eloise-nebula: I think it would be nice to include the datestr here. It would be useful for matching caters to their respective errors.

lcls_tools/common/logger/logger.py
```diff
@@ -1,23 +1,38 @@
-import logging
 import sys
+import logging
+
+
+FORMAT_STRING = "%(asctime)s - %(levelname)s - %(message)s"
+DATE_FORMAT = "%H:%M:%S"
```
eloise-nebula, line 6: With my previous comment in mind, does it make sense to include the date too? Or could we generate logfiles by session?

lcls_tools/common/logger/logger.py
```diff
@@ -1,23 +1,38 @@
...
-        out_handler = logging.StreamHandler(sys.stdout)
-        out_handler.setLevel(logging.DEBUG)
-        out_handler.setFormatter(formatter)
+    # Exception hook
+    def handle_exception(exc_type, exc_value, exc_traceback):
+        if issubclass(exc_type, KeyboardInterrupt):
+            sys.__excepthook__(exc_type, exc_value, exc_traceback)
+            return
+        logger.error(
+            "Unhandled exception", exc_info=(exc_type, exc_value, exc_traceback)
```
eloise-nebula, line 34: In a future update we should make this handler actually handle errors per GUI. Oblique errors like this one result in a lot of confused tickets.

eloise-nebula (approved):
Thanks for the clarification. It looks good in general, but I think we'd like more date information somewhere in the logs. Things to think about in the future include:
-  GUI-side error handling
- Will a log file persist per session? per day?

I'm approving since it looks close to done. Thank you!

lcls_tools/common/logger/logger.py
kabanaty: The date is included with `%(asctime)s`.  The argument `datefmt` determines whether the custom `DATE_FORMAT` or the default is used.  

The File Handler entries will look like this: `07/09/2025 03:42:17 PM - INFO - Test message`
The Console Handler entries will look like this: `15:42:17 - INFO - This is a test message.`  I figured that the console doesn't need to get bogged down with the full datestring if the user is reading it straight from the CLI.

Log files could absolutely be generated by session/day/arbitrary time period as long as the appropriate `log_file` is passed to the logger object.

PR #312, opened 2025-07-25
Update DIAG0.yaml

roussel-ryan:
add XCDGTCX / YCDGTCX to DIAG0

### DIFF
--- a/lcls_tools/common/devices/yaml/DIAG0.yaml
+++ b/lcls_tools/common/devices/yaml/DIAG0.yaml
@@ -1,5 +1,5 @@
 bpms:
-  BPMDG000?:
+  BPMDG000:
     controls_information:
       PVs: {}
       control_name: BPMS:DIAG0:136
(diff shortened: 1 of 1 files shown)


nneveu:
Can the writer script be updated to add this instead of a manual change? Are we not able to pull this w/ meme?

roussel-ryan:
Good point, although I'm not sure how to do this, maybe @phys-cgarnier can help

kabanaty:
Do we know what ‘*DIAG0’ indicates vs ‘DIAG0’? We should discuss here or at the next group meeting about refreshing the .csv file. I believe it’s currently over a year out of date. Also, if we need to adjust details (like area names) we can contact Andrea Chan or Mark Woodley.

eloise-nebula:
I asked Matt Gibbs about this notation earlier this year. He said, "The asterisk notes that this is a "Deferred" item - it was planned, but got cut (usually for budget reasons)." Chances are if there are installed devices that are still under the starred area that just means that Oracle wasn't updated properly.

PR #315, opened 2025-08-07
Wire scan bug fixes and reliability improvements

kabanaty:
This pull request includes a set of improvements and bug fixes identified during the final MD shift of the run and subsequent testing during the summer shutdown.

**Changes**
- Added Gaussian fit support for profile analysis
- Integrated a custom logging module for consistent and configurable log output
- Implemented checks for wire enable status prior to scan execution
- Expanded and standardized logging across key components
- Improved error handling for greater robustness
- Introduced utility functions `active_profiles()` and `calc_buffer_points()`
- Updated expected data type for `MeasurementResult` to ensure consistency

### DIFF
--- a/lcls_tools/common/measurements/wire_scan.py
+++ b/lcls_tools/common/measurements/wire_scan.py
@@ -281,19 +281,19 @@ class WireBeamProfileMeasurement(Measurement):
 
     def get_profile_range_indices(self):
         """
-        Finds sequential scan indices within each plane's position range.
+        Finds sequential scan indices within each profile's position range.
 
         Filters wire position data to identify index ranges for x, y, and u
-        planes, excluding non-continuous points like wire retractions.
+        profiles, excluding non-continuous points like wire retractions.
 
         Returns:
-            dict: Plane keys ('x', 'y', 'u') with lists of index arrays.
+            dict: Profile keys ('x', 'y', 'u') with lists of index arrays.
         """
         self.logger.info("Getting profile range indices...")
-        # Get wire data to detemine plane indices
+        # Get wire data to detemine profile indices
         position_data = self.data[self.my_wire.name]
 
-        # Hold plane ranges
+        # Hold profile ranges
         ranges = {}
 
         # Hold sequential indices (avoid catching return wires)
@@ -337,13 +337,13 @@ class WireBeamProfileMeasurement(Measurement):
 
     def organize_data_by_profile(self, profile_idxs):
         """
-        Organizes detector data by scan plane for each device.
+        Organizes detector data by scan profile for each device.
 
         Uses sequential indices to separate full device data into
-        x, y, and u plane datasets.
+        x, y, and u profile datasets.
 
         Returns:
-            dict: Nested dict with planes as keys and device data per plane.
+            dict: Nested dict with profiles as keys and device data per profile.
         """
         self.logger.info("Creating profile data objects...")
         profiles = list(profile_idxs.keys())
@@ -374,13 +374,13 @@ class WireBeamProfileMeasurement(Measurement):
 
     def fit_data_by_profile(self):
         """
-        Fits detector data for each plane and device.
+        Fits detector data for each profile and device.
 
         Applies beam fitting to x, y, and u projections
         for all devices in the detector data.
 
         Returns:
-            dict: Fit results organized by plane and device.
+            dict: Fit results organized by profile and device.
         """
         self.logger.info("Fitting profile data...")
         # Get list of profiles from data set
--- a/lcls_tools/common/measurements/wire_scan.py
+++ b/lcls_tools/common/measurements/wire_scan.py
@@ -241,7 +241,7 @@ class WireBeamProfileMeasurement(Measurement):
         self.logger.info(
             "BSA buffer %s acquisition complete after %s seconds",
             self.my_buffer.number,
-            i/10,
+            i / 10,
         )
 
         self.logger.info("Getting data from BSA buffer...")
@@ -270,7 +270,7 @@ class WireBeamProfileMeasurement(Measurement):
                     for lblm in self.my_wire.metadata.lblms
                 }
             )
-        self.logger.info("Data retrieved from BSA buffer.")
+        self.logger.info("Data retrieved from BSA buffer.  Scan complete.")
 
         # Release EDEF/BSA
         self.logger.info("Releasing BSA buffer.")
@@ -306,6 +306,11 @@ class WireBeamProfileMeasurement(Measurement):
             method_name = f"{p}_range"
             ranges[p] = getattr(self.my_wire, method_name)
 
+            if position_data.max() < ranges[p][0]:
+                msg = f"Scan did not reach expected {p} profile range.  Exiting scan."
+                self.logger.error(msg)
+                raise RuntimeError(msg)
+
             # Get indices of when position is within a selected wire scan profile range
             idx = np.where(
                 (position_data >= ranges[p][0]) & (position_data <= ranges[p][1])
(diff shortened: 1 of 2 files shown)


lcls_tools/common/measurements/wire_scan.py
```diff
@@ -214,33 +299,40 @@ def get_profile_range_indices(self):
...
+                (position_data >= ranges[p][0]) & (position_data <= ranges[p][1])
             )[0]
 
+            # Data slice representing a given profile measurement
             pos = position_data[idx]
+
+            # Boolean mask of monotonically non-decreasing data points
+            # Mask of values where difference between neighbors is positive or zero
             mono_mask = np.diff(pos) >= 0
             mono_mask = np.concatenate(([True], mono_mask))
```
eloise-nebula, line 320: In the following case you can get monotonically decreasing data:
```
>>> a = np.array([10, 8, 9, 7, 8, 6, 7]);
>>> i = np.concatenate(([True], np.diff(a) >= 0))
>>> a[i]
array([10,  9,  8,  7])
```
If this case is acceptable for the wire scanner then maybe the comment can be reworded. Otherwise maybe another solution is necessary?

lcls_tools/common/measurements/wire_scan.py
```diff
@@ -214,33 +299,40 @@ def get_profile_range_indices(self):
...
-            profile_idxs[profile] = mono_idx
+            # Boolean mask of indices in a given profile measurement
+            try:
+                mono_idx = idx[mono_mask]
+                profile_idxs[p] = mono_idx
+            except Exception as e:
+                msg = (
+                    f"Could not determine '{p}' profile data range. "
+                    "This usually means the wire didn't move properly or the "
+                    "buffer did not properly capture the motion data."
```
eloise-nebula, line 330: Is there a way to differentiate these two errors? This also seems like something the user would want to know after `scan_with_wire` or `get_data_from_bsa`.

lcls_tools/common/measurements/wire_scan.py
```diff
@@ -165,20 +229,39 @@ def get_data_from_bsa(self):
             dict: Collected data keyed by device name.
         """
         data = {}
+
         # Wait for buffer 'ready'
+        i = 1
         while not self.my_buffer.is_acquisition_complete():
```
eloise-nebula, line 235: It's possible to abort the scan during this loop using the abort_scan pv. Is there any way to check if that's happened here and throw a unique error?

eloise-nebula, line 235: It seems to me like the user would always want to acquire buffer data after initiating a scan. Should this function be combined with the one above?

eloise-nebula (commented):
Hi Tyler, thanks for the PR. The code reads easily. Right now the code doesn't handle the case where the wire scan is aborted, which could lead to some obscure/unexpected errors. I'm not clear on why `get_data_from_bsa` is a distinct function from `scan_with_wire`, especially since the code that waits for the wire scan to complete is in `get_data_from_bsa`.

lcls_tools/common/measurements/wire_scan.py
```diff
@@ -135,19 +163,55 @@ def scan_with_wire(self):
...
+            time.sleep(0.1)
+
+        self.logger.info(
+            "%s initialized after %s seconds", self.my_wire.name, elapsed_time
+        )
 
         # Start buffer
-        print("Starting BSA buffer...")
+        self.logger.info("Starting BSA buffer...")
         self.my_buffer.start()
```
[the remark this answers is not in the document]
kabanaty: Right now there's no formal indication that a scan has completed.  The two possibilities that jump out at me could be 1) when the BSA buffer is ready to read, or 2) when the wire returns to the 'homed' (fully retracted) position.  I prefer option 1 because it's possible the wire could get stuck and not re-home.  Making a logging call that states "Scan complete" would be informative for the users.

lcls_tools/common/measurements/wire_scan.py
```diff
@@ -125,7 +142,18 @@ def create_device_dictionary(self):
                     region=self.my_wire.area,
                 )
             else:
-                devices[name] = create_lblm(area=area, name=name)
+                try:
+                    # Make LBLM object
+                    devices[name] = create_lblm(area=area, name=name)
+                except ValidationError as e:
+                    self.logger.warning(
+                        "Validation error creating %s. Continuing...", name
```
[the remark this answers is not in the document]
kabanaty: During the LBLM re-commissioning certain devices are unavailable during repairs.  As they come on and off-line I'm finding that it's cleaner to log this warning message instead of the full Pydantic error stack.  There's no bug to fix, just managing the changing device landscape.

I think moving forward this is an acceptable solution.  A particular device validation error will be logged and the diagnosis will happen in the `create_lblm` function separately.

lcls_tools/common/measurements/wire_scan.py
kabanaty: Oh excellent catch.  I've been struggling to find a decent solution for this for a while.  I haven't got caught with this particular case yet.
I could include something like:

```suggestion
            mono_mask = np.diff(pos) >= 0
            mono_mask = np.concatenate(([True], mono_mask))
            
            # Find first false
            bad = np.where(~mono_mask)
            if bad.size:
                end = bad[0]                # stop before the first decrease
                mono_idx = idx[:end] 
            else:
                mono_idx = idx            # indices were monotonically non-decreasing to start with
```

lcls_tools/common/measurements/wire_scan.py
kabanaty: This exception arises because either the wire did not insert fully into an expected profile range (ex. it wants to scan from 39-46mm but failed and retracted after 30mm) or the buffer did not collect some or all of the position data (ex. buffer started early or late and did not get data, or did not collect any data at all).

The wire positions get printed to the log (see line 238) so the first failure mode can diagnosed from the log file.
If I'm expecting to scan X, Y, and U profiles, I could do a check to compare that the wire makes it to the minimum position for the expected ranges.  That should catch wire motion errors.

The usual failure mode for BSA presents as an array of identical values.  That would be trivial to check, and if the buffer starts out-of-sync with the device, then that should present itself as the first kind of error anyways.  I'll work on a fix.

lcls_tools/common/measurements/wire_scan.py
kabanaty: That is correct, the user could abort scan during this loop.  What actually happens during an abort scan call is the `MOTR.STOP` PV gets written to which stops the motor in place.  I don't think there is any indication from the controls system that the wire scanner is put into this state.  The BSA buffer will continue as usual and an error will likely be thrown later on when the profile ranges are analyzed.

lcls_tools/common/measurements/wire_scan.py
kabanaty: This looks like a really good solution! Thanks, Eloise.  I've implemented this in the code slightly differently, to preserve `mono_mask` and use that to isolate the profile indices.

lcls_tools/common/measurements/wire_scan.py
eloise-nebula: I missed that. Thanks for the clarification!

eloise-nebula: That's unfortunate. No changes necessary for now, but we should think about this and maybe talk to EED in the future.

shamin-slac:
Hi, sorry I'm late on reviewing this. I think these are great changes! You're making good use of the severity levels for the logging, and I agree with the error handling you've added such as with the lblm creation and the timeout in the wire initialization. I'll hold off on approving so you can put in the fix for distinguishing between the wire motion error and the buffer data collection error.

eloise-nebula (approved):
New changes look good. Thank you!

PR #317, opened 2025-08-27
yaml.write now supports multiple areas as intended.

eloise-nebula:
Fixed long standing bug in yaml_writer that broke write if called with multiple areas. I swore I fixed this multiple times. Hopefully this PR will put it to rest.


kabanaty (approved):
Looks good! Thanks for the quick fix.

PR #318, opened 2025-08-28
Add functions for saving/loading measurement results as h5 files

shamin-slac:
I imagine a bunch of us basically have something like this written individually, so I thought I'd add them here. These are functions that allow one to save/load a measurement result object (e.g. EmittanceMeasurementResult) as an h5 file. While these were written with measurement result objects in mind, they're actually quite general, so I'm open to modifying them to that effect.

- [ ] Add test for saving/loading

### DIFF
--- a/lcls_tools/common/io.py
+++ b/lcls_tools/common/io.py
@@ -1,25 +1,64 @@
-from typing import Optional
+import json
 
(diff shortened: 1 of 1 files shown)


shamin-slac:
@eloise-nebula I ran the tests again after updating #325, and they're all passing. So I think it should be good.

PR #328, opened 2025-09-29
move validation and serialization of PV objects to base PVSet

roussel-ryan:
This pull request refactors the way PV (Process Variable) field validation and serialization are handled for device PVSet classes. The main change is centralizing PV string-to-object conversion and serialization logic in the base `PVSet` class, eliminating duplicated code in each device-specific subclass. This results in cleaner, more maintainable code.

**Core PVSet validation and serialization improvements:**

* Added a `field_validator` to the base `PVSet` class in `device.py` to convert PV strings to `PV` objects for all subclasses, and a `field_serializer` to convert `PV` objects back to strings. This ensures consistent handling of PV fields across all device types.

**Code deduplication and cleanup across device modules:**

* Removed redundant `field_validator` methods from all device-specific `PVSet` subclasses (`BPMPVSet`, `ICTPVSet`, `LBLMPVSet`, `MagnetPVSet`, `PMTPVSet`, `ScreenPVSet`, `TCAVPVSet`, `WirePVSet`), since the logic is now handled in the base class. [[1]](diffhunk://#diff-a0b51b8b22e91b19adb536859d14c3f7f312d1ecf5a298b951e82a5e5e5c30d2L31-L34) [[2]](diffhunk://#diff-9881a64bbeea86670d91c8287292b033401d27c7bceea7ad155690e6a22de444L19-L23) [[3]](diffhunk://#diff-430b5b9b571d9e74bfe6287f8c7bfd74f3d4ee97e2eb6f9016acfd0a06cabc04L47-L50) [[4]](diffhunk://#diff-443ce4965bfca0dd51eaa77b8d03790820b442edcd08b034df67279aa639918eL38-R38) [[5]](diffhunk://#diff-76f6e9af90cfd0d54572696dbbad84781389046f06335fd403c5ef4e7bc9859bL27-L30) [[6]](diffhunk://#diff-cb99364424bfec4fa3b739fc682ce88928c90419f5ba5dcec7f3df4ae13ece53L58-L62) [[7]](diffhunk://#diff-c8c7f0561b29c7165d1e6e206f3cc7fb69642e5f8526e593f674e15738d979b7L32-L35) [[8]](diffhunk://#diff-08b0948a36bc9d6b7568850785e9bfa959f9c24a6746a3bb10bb497b61edf93fL93-L96)

### DIFF
--- a/lcls_tools/common/devices/device.py
+++ b/lcls_tools/common/devices/device.py
@@ -14,6 +14,8 @@ class PVSet(lcls_tools.common.BaseModel):
 
     @field_validator("*", mode="before")
     def validate_pv_fields(cls, v: str) -> PV:
+        if v is None:
+            return None
         return PV(v)
 
     @field_serializer("*")
--- a/lcls_tools/common/devices/device.py
+++ b/lcls_tools/common/devices/device.py
@@ -18,6 +18,8 @@ class PVSet(lcls_tools.common.BaseModel):
 
     @field_serializer("*")
     def serialize_pv_fields(self, v: PV, _info) -> str:
+        if v is None:
+            return None
         return v.pvname
 
 
--- a/lcls_tools/common/devices/device.py
+++ b/lcls_tools/common/devices/device.py
@@ -1,4 +1,4 @@
-from pydantic import SerializeAsAny, ConfigDict, field_validator
+from pydantic import SerializeAsAny, ConfigDict, field_validator, field_serializer
 from typing import List, Union, Callable, Dict, Optional
 from epics import PV
 
@@ -11,7 +11,16 @@ class PVSet(lcls_tools.common.BaseModel):
         extra="forbid",
         frozen=True,
     )
-    ...
+
+    @field_validator("*", mode="before")
+    def validate_pv_fields(cls, v: str) -> PV:
+        return PV(v)
+
+    @field_serializer("*")
+    def serialize_pv_fields(self, v: PV, _info) -> str:
+        if v is None:
+            return None
+        return v.pvname
 
 
 class ControlInformation(lcls_tools.common.BaseModel):
--- a/lcls_tools/common/devices/device.py
+++ b/lcls_tools/common/devices/device.py
@@ -14,14 +14,10 @@ class PVSet(lcls_tools.common.BaseModel):
 
     @field_validator("*", mode="before")
     def validate_pv_fields(cls, v: str) -> PV:
-        if v is None:
-            return None
         return PV(v)
 
     @field_serializer("*")
     def serialize_pv_fields(self, v: PV, _info) -> str:
-        if v is None:
-            return None
         return v.pvname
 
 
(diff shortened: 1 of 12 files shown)


lcls_tools/common/devices/device.py
```diff
@@ -11,7 +11,18 @@ class PVSet(lcls_tools.common.BaseModel):
         extra="forbid",
         frozen=True,
     )
-    ...
+
+    @field_validator("*", mode="before")
+    def validate_pv_fields(cls, v: str) -> PV:
+        if v is None:
+            return None
```
eloise-nebula, line 18: This if statement is not necessary. Default values bypass the field validator, and non-optional fields can't can't be set to None. For instance, this code block runs without error:
```
import pydantic
from typing import Optional

class Foo(pydantic.BaseModel):
    doe: Optional[str] = None
    ray: str


    @pydantic.field_validator("doe", mode="before")
    def validate_doe(cls, v: str) -> str:
        assert(v is not None)
        return repr(v)

    @pydantic.field_validator("ray", mode="before")
    def validate_ray(cls, v: any) -> str:
        if v is None:
            return None
        return repr(v)

f = Foo(ray='') # validate_doe is not called
assert(f.doe is None) # so f.doe can be None

try:
    f = Foo(doe=None, ray='') # validate_doe is called
except pydantic.ValidationError as e:
    # we got an assertion error
    assert(e.errors()[0]['type'] == "assertion_error")

try:
    f = Foo(ray=None) # validate_ray is called
except pydantic.ValidationError as e:
    # input needs to be a string
    assert(e.errors()[0]['type'] == "string_type")
```

lcls_tools/common/devices/device.py
```diff
@@ -11,7 +11,18 @@ class PVSet(lcls_tools.common.BaseModel):
...
+        return PV(v)
+
+    @field_serializer("*")
+    def serialize_pv_fields(self, v: PV, _info) -> str:
+        if v is None:
+            return None
```
eloise-nebula, line 24: This if statement is also unnecessary.
```
class Foo(pydantic.BaseModel):
    doe: Optional[str] = None

f = Foo()
assert(f.model_dump() == {'doe': None})
```

eloise-nebula (changes requested):
Hi Ryan, thanks for the PR. A couple of minor comments. Looks good overall.

lcls_tools/common/devices/device.py
roussel-ryan: so the validation one is not necessary, but this one is when using the PV objects since the code below tried to get a property of `v` which will fail

lcls_tools/common/devices/device.py
roussel-ryan: You can pass None to optional fields

roussel-ryan: if we want to serialize the entire object, that would include fields that have None values

robnagler: I see. Thanks.

eloise-nebula (approved):
Thanks for the changes. Looks good to me.

PR #370, opened 2026-05-19
Add minimum beam size check for fit validation.

dylanmkennedy:
Adds the argument 'min_beam_size' to the ImageProjectionFit class. This argument specifies the minimum size (in pixels) of an acceptable projection fit. When 'validate_fit' is True, fit results with sizes below this threshold will be set to NaN.


kabanaty (approved):
Matches `slac-measurements` PR

PR #371, opened 2026-05-19
Adds option to weight nonlinear fitting in emittance calc.

dylanmkennedy:
Adds boolean argument 'weighted_fit' specifying whether or not to weight the nonlinear fitting in emittance calc by inverse beam size. If True, the nonlinear fitting will prioritize smaller beamsizes.


kabanaty (approved):
Discussed slac-tools meeting 6/22

